PR #108, opened 2022-01-20
[CN-262] Add Kubernetes provider info into platform struct

mtyazici:
Changes:
- Added provider info to Platform struct, providers are `EKS`, `GKE` and `AKS`.
- Added config to the Platform resolver client. It used to use Kubeconfig as default one, I set its config to the Manager's one.

### DIFF
--- a/controllers/platform/platform.go
+++ b/controllers/platform/platform.go
@@ -22,7 +22,7 @@ const (
 type ProviderType string
 
 const (
-	AWS ProviderType = "AWS"
+	EKS ProviderType = "EKS"
 	AKS ProviderType = "AKS"
 	GKE ProviderType = "GKE"
 )
@@ -133,7 +133,7 @@ func getInfo() (Platform, error) {
 	for _, v := range apiList.Groups {
 		switch v.Name {
 		case "crd.k8s.amazonaws.com":
-			info.Provider = AWS
+			info.Provider = EKS
 			return info, nil
 		case "networking.gke.io":
 			info.Provider = GKE
--- a/controllers/platform/platform.go
+++ b/controllers/platform/platform.go
@@ -5,6 +5,12 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/client/config"
 )
 
+type Platform struct {
+	Type     PlatformType `json:"type"`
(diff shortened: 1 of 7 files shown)


controllers/platform/platform.go
```diff
@@ -5,17 +5,26 @@ import (
...
 )
 
-type Platform struct {
-	Type    PlatformType `json:"type"`
-	Version string       `json:"version"`
-}
+type ProviderType string
+
+const (
+	AWS ProviderType = "AWS"
```
hasancelik, line 24: It should be `EKS`

hasancelik (changes requested):

controllers/platform/platform.go
mtyazici: Fixed it, thanks

controllers/platform/platform.go
```diff
@@ -29,11 +44,24 @@ func GetPlatform() (Platform, error) {
 	var err error
 	plt, err = getInfo()
 	if err != nil {
-		return plt, err
```
hakanmemisoglu, line 32: I'd convert this to return `nil` and change the function signature.

hakanmemisoglu (approved):
LGTM. Just a small comment...

controllers/platform/platform.go
mtyazici: I agree with returning nil but what do you mean by changing function signature? Do you mean returning pointers?

mtyazici:
@hakanmemisoglu and @hasancelik,

I am thinking of setting the platform struct only at the operator start up. The struct will be set and consequent `get`s only will return the field. However, this requires that if the platform information cannot be set at startup the operator will exit.

The only reason I did it this way is because the Kubernetes version could get updated and operator would not be aware of that. However, operator will be restarted when there is a Kubernetes version update. So getting the Platform information only at startup is enough. WDYT?

hasancelik:
@mtyazici 

Good approach 👍  The operator has to retrieve Platform(Kubernetes | OCP) information during start-up because we modify our resources according to this info. So exit scenario should not be a problem for us.

hasancelik (approved):

PR #112, opened 2022-01-25
[CN-276] Configuration of persistence feature for members

SeriyBg:

PR #114, opened 2022-01-31
[CN-276] add HotBackup CR and implementation

SeriyBg:
- Add `HotBackup` CR
- Add `HotBackup` triggering implementation using `REST API`
- Fix small issue with default value for `ClusterName`


SeriyBg:
This PR is based on #112 so it's better to review it after it's merged to `main`

controllers/hazelcast/hot_backup_controller.go
```diff
@@ -0,0 +1,76 @@
...
+			logger.Info("HotBackup resource not found. Ignoring since object must be deleted")
+			return ctrl.Result{}, nil
+		}
+		logger.Error(err, "Failed to get HotBackup")
+		return ctrl.Result{}, err
+	}
+	h := &hazelcastv1alpha1.Hazelcast{}
+	err = r.Client.Get(ctx, types.NamespacedName{Namespace: req.Namespace, Name: hb.Spec.HazelcastName}, h)
+	if err != nil {
+		logger.Error(err, "Could not trigger Hot Backup: Hazelcast resource not found")
```
hasancelik, line 49: When I try to create HotBackup CR with a non-existing HZ CR name, the operator returned an error but it still tries to change the cluster state:
```
2022-02-14T17:03:46.179+0300    ERROR   controllers.HotBackup   Could not trigger Hot Backup: Hazelcast resource not found      {"hazelcast-hot-backup": "default/hot-backup", "error": "Hazelcast.hazelcast.com \"test\" not found"}
github.com/go-logr/zapr.(*zapLogger).Error
.....
2022-02-14T17:03:46.183+0300    ERROR   controllers.HotBackup   Error creating HotBackup. Could not change the cluster state to PASSIVE {"hazelcast-hot-backup": "default/hot-backup", "error": "Post \"http://..svc.cluster.local:5701/hazelcast/rest/management/cluster/changeState\": dial tcp: lookup ..svc.cluster.local: no such host"}
```

api/v1alpha1/hazelcast_types.go
```diff
@@ -278,8 +278,40 @@ type HazelcastList struct {
...
 
+// HotBackupStatus defines the observed state of HotBackup
+type HotBackupStatus struct {
+}
+
+// HotBackupSpec defines the Spec of HotBackup
+type HotBackupSpec struct {
+	// HazelcastName defines the name of the Hazelcast resource
+	// +kubebuilder:validation:Required
+	HazelcastName string `json:"hazelcastName"`
```
hasancelik, line 289: @SeriyBg I am not sure but `hazelcastName` might be confusing for users. probably we will encounter this question multiple times --> Is it cluster-name? :) I do not have a strong candidate to replace, `resourceName` or `crName`might be better, I think. If the user is applying `HotBackup` CR, we can assume that he/she already understands the context.

controllers/hazelcast/hazelcast_resources.go
```diff
@@ -726,6 +729,9 @@ func (r *HazelcastReconciler) applyDefaultHazelcastSpecs(ctx context.Context, h
 		h.Spec.ClusterSize = n.DefaultClusterSize
 		changed = true
 	}
+	if h.Spec.ClusterName == "" {
```
hasancelik, line 732: 👍

hasancelik (changes requested):

api/v1alpha1/hazelcast_types.go
SeriyBg: @hasancelik agreed, with one addition that I would keep the `hazelcast` word in it, like `hazelcatResourceName`

hasancelik: `hazelcatResourceName` sounds good to me 👍

controllers/hazelcast/hot_backup_controller.go
SeriyBg: Oops, missing `return` statement

api/v1alpha1/hazelcast_types.go
```diff
@@ -278,8 +278,40 @@ type HazelcastList struct {
 	Items           []Hazelcast `json:"items"`
 }
 
+// HotBackupStatus defines the observed state of HotBackup
```
mtyazici, line 281: Is there a reason why we did not create a new file for the HotBackup CR?

mtyazici (approved):
LGTM, only put one question.

api/v1alpha1/hazelcast_types.go
hasancelik: Separate `_types.go` for every CR is a good idea, I think 👍

SeriyBg: For no specific reason, I've decided to put it in the same file based on the other operators' code structure. But I will move it to the separate file

controllers/hazelcast/hot_backup_controller.go
```diff
@@ -0,0 +1,77 @@
+package hazelcast
```
mtyazici, line 1: I think this should also be split into `hotbackup` folder.

SeriyBg: We can split it, but `hotbackup` CR is directly related to the `hazelcast` one. That's why I think it's better to keep it inside the `hazelcast` package. Wdyt?

controllers/hazelcast/hot_backup_controller.go
```diff
@@ -0,0 +1,77 @@
...
+		}
+		logger.Error(err, "Failed to get HotBackup")
+		return ctrl.Result{}, err
+	}
+	h := &hazelcastv1alpha1.Hazelcast{}
+	err = r.Client.Get(ctx, types.NamespacedName{Namespace: req.Namespace, Name: hb.Spec.HazelcastResourceName}, h)
+	if err != nil {
+		logger.Error(err, "Could not trigger Hot Backup: Hazelcast resource not found")
+		return ctrl.Result{}, err
+	}
```
mtyazici, line 51: Can we also put a step for checking if the Hazelcast Custom Resource is in a ready state?

controllers/hazelcast/hot_backup_controller.go
mtyazici: I agree it is directly related but it is also a different CR and has a separate controller, to be honest I am not sure what is the best way to handle it.

mtyazici: I am OK with the current status.

hasancelik (approved):

PR #122, opened 2022-02-07
Fix annotation for expose externally member access

### LINKED ISSUE #95: Wrong annotation when using Smart client with no MemberAccess defined.
Default value for `memberAccess` is `NodePortExternalIP`. However when I supply the following Hazelcast CR:

```yaml
kind: Hazelcast
metadata:
  name: hazelcast
spec:
  clusterSize: 3
  repository: 'docker.io/hazelcast/hazelcast-enterprise'
  version: '5.0'
  licenseKeySecret: hazelcast-license-key
  exposeExternally:
    type: Smart
    discoveryServiceType: LoadBalancer
```

The annotation for the memberAccess in Hazelcast pods becomes 

```yaml
metadata:
  annotations:
    hazelcast.com/expose-externally-member-access: ""
```

We need to set the default value for the memberAccess when Smart client is enabled.

mtyazici:
Fixes #95.

### DIFF
--- a/api/v1alpha1/hazelcast_types.go
+++ b/api/v1alpha1/hazelcast_types.go
@@ -150,7 +150,7 @@ func (c *ExposeExternallyConfiguration) DiscoveryK8ServiceType() corev1.ServiceT
 	}
 }
 
-// Returns service type that is used for the communication with each member (NodePort by default).
+// Returns the member access type that is used for the communication with each member (NodePortExternalIP by default).
 func (c *ExposeExternallyConfiguration) MemberAccessType() MemberAccess {
 	if c.MemberAccess != "" {
 		return c.MemberAccess
--- a/api/v1alpha1/hazelcast_types.go
+++ b/api/v1alpha1/hazelcast_types.go
@@ -151,6 +151,14 @@ func (c *ExposeExternallyConfiguration) DiscoveryK8ServiceType() corev1.ServiceT
 }
 
 // Returns service type that is used for the communication with each member (NodePort by default).
+func (c *ExposeExternallyConfiguration) MemberAccessType() MemberAccess {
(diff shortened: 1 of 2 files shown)


api/v1alpha1/hazelcast_types.go
```diff
@@ -150,6 +150,14 @@ func (c *ExposeExternallyConfiguration) DiscoveryK8ServiceType() corev1.ServiceT
 	}
 }
 
+// Returns service type that is used for the communication with each member (NodePort by default).
+func (c *ExposeExternallyConfiguration) MemberAccessType() MemberAccess {
+	if c.MemberAccess != "" {
+		return c.MemberAccess
+	}
+	return MemberAccessNodePortExternalIP
```
hakanmemisoglu, line 158: We can also set the default for the field here.

hakanmemisoglu (approved):

api/v1alpha1/hazelcast_types.go
mtyazici: It will only be held in the memory though, will not persist anywhere. I tried to add the defaults in the `applyDefaultHazelcastSpecs` function but I had to add 3 nested if statements, so I dropped it :D.

SeriyBg: @mtyazici I might be missing something, but couldn't we just check for `h.Spec.ExposeExternally.IsEnabled() && h.Spec.ExposeExternally.MemberAccess == ""` for the default config?

mtyazici: Also we need to know if Smart client is enabled. I did not just want to add default values for `memberAccess` but for all fields in exposeExternally.

SeriyBg (approved):

PR #126, opened 2022-02-10
[CN-308] Add existingVolumeClaimName into MC persistence

mtyazici:
Other Changes:
- When we apply empty config, persistence was not enabled even though we have true as the default `enabled` value. We also needed to give default value for the cases  where persistence object was empty or not present.


api/v1alpha1/managementcenter_types.go
```diff
@@ -82,12 +83,20 @@ const (
 type PersistenceConfiguration struct {
 	// +optional
 	// +kubebuilder:default:=true
+	// When true MC will use a PersistentVolumeClaim to store data.
```
SeriyBg, line 86: It's a bit of nitpicking but seems like we always put the docs as the first line of the comment and the `kubebuilder` annotations below it

test/integration/managementcenter_controller_test.go
```diff
@@ -157,4 +158,39 @@ var _ = Describe("ManagementCenter controller", func() {
...
+				assertExists(lookupKey(fetchedCR), fetchedSts)
+				expectedVolume := corev1.Volume{
+					Name: n.MancenterStorageName,
+					VolumeSource: corev1.VolumeSource{
+						PersistentVolumeClaim: &corev1.PersistentVolumeClaimVolumeSource{
+							ClaimName: "ClaimName",
+						},
+					},
+				}
+				Expect(fetchedSts.Spec.Template.Spec.Volumes).To(ContainElement(expectedVolume))
```
SeriyBg, line 185: Since it is expected only one `Volume` to be present I would suggest using `ConsistOf` instead of `ContainsElement` as this is a more narrow check

SeriyBg (approved):
LGTM, I've added two nitpicking comments though 😄

test/integration/managementcenter_controller_test.go
mtyazici: Will change it to that.

controllers/managementcenter/managementcenter_resources.go
```diff
@@ -287,7 +286,11 @@ func (r *ManagementCenterReconciler) reconcileStatefulset(ctx context.Context, m
 
 	if mc.Spec.Persistence.IsEnabled() {
 		sts.Spec.Template.Spec.Containers[0].VolumeMounts = []v1.VolumeMount{persistentVolumeMount()}
-		sts.Spec.VolumeClaimTemplates = append(sts.Spec.VolumeClaimTemplates, persistentVolumeClaim(mc))
+		if mc.Spec.Persistence.ExistingVolumeClaimName == "" {
+			sts.Spec.VolumeClaimTemplates = []corev1.PersistentVolumeClaim{persistentVolumeClaim(mc)}
+		} else {
+			sts.Spec.Template.Spec.Volumes = []v1.Volume{existingVolumeClaim(mc.Spec.Persistence.ExistingVolumeClaimName)}
```
hakanmemisoglu, line 292: Checking the access mode of the PVC would be a good guard.

hakanmemisoglu (approved):
It looks good. I just dropped one comment.

controllers/managementcenter/managementcenter_resources.go
mtyazici: Do you mean we should give operator read access to PVCs and check if the given PVC is writable?

hakanmemisoglu: Yes, that's what I mean. If we don't have GET access for PVCs, then we can add it later. It is not so important. 
The only problem is that we will not catch this error until MC tries to write something to disk.

mtyazici: Is there such a access mode?

mtyazici: Guys, I checked to see why PVCs also have multiple access modes and could not find an answer. Do you know why someone would put multiple access modes into a PVC, and what it would achieve?

SeriyBg: > Is there such an access mode? It seems there is, but it will be included in v1.22+

But event with `ReadWriteOnce`, wouldn't it be a problem if operator and MC pods will be deployed on different nodes?

hakanmemisoglu: @SeriyBg I don't understand your point. What I am trying to say is that check whether PVC has write access or read only. If MC uses underlying disk or filesystem to write something and if it is read only volume, then it might fail at runtime. ReadWriteOnce, ReadWriteMany, ReadWriteOncePod are OK values. ReadOnlyMany is the only problematic value.

@mtyazici As far as I understand when you specifiy PVC with set of A access modes and a PV with set of B access modes. PVC will be bound to the PV if and only if A &SubsetEqual; B.

SeriyBg: @hakanmemisoglu oh so you mean just to check that it has the right `accessModes`? That sounds reasonable

mtyazici: But which of the access modes will be enabled though? A PVC can only use one right?

hakanmemisoglu: volumeMount.readOnly will define it. In theory, when I create a PVC, it means that I want a volume which support access modes ReadWriteOnce, ReadWriteMany etc. If it is dynamically provisioned, then the cluster will create a matched PV for it, specifiying all access modes (and it can be all of them, for example cloud provisioner will create a volume with NFS). How you mount the volume to the container is different thing. 

By the way, I tried to demonstrate a volume with multiple access modes, GKE creates it. However, you can not mount it multiple pods as read only. I am trying to understand why. It seems GKE disk is mounted to one node and a pod from another cannot mount it. It is weird. It is doing the same thing even though I create a PVC with only ReadOnlyMany access.

As I said before, it is just a small guard. It is not important. It seems that even the implementation (GKE) lies about it.

PR #127, opened 2022-02-11
[CN-258] Implement External IP status field for HZ and MC

hakanmemisoglu:
Adds status field `externalAddresses` to Hazelcast and ManagementCenter CRs.

The field will be showed in `kubectl get` output. 
External addresses will not be empty in case:
- Hazelcast is configured with ExposeExternally Discovery type LoadBalancer.
- ManagementCenter is configured with ExposeExternally LoadBalancer type.


controllers/util/util.go
```diff
@@ -171,3 +173,62 @@ func deploymentName(podName string) string {
 	s := strings.Split(podName, "-")
 	return strings.Join(s[:len(s)-2], "-")
 }
+
+func GetLoadBalancerAddress(lb *corev1.LoadBalancerIngress) string {
```
SeriyBg, line 177: Can this func be private?

controllers/util/util.go
```diff
@@ -171,3 +173,62 @@ func deploymentName(podName string) string {
...
+
+	result := strings.Join(externalAddrs, ",")
+	if result == "" {
+		logger.Info("LoadBalancer external IP is not ready")
+		return ""
+	}
+	return result
+}
+
+func GetDiscoveryService(ctx context.Context, cli client.Client, cr Predefined) (*corev1.Service, error) {
```
SeriyBg, line 226: Also this one

controllers/util/util.go
```diff
@@ -171,3 +173,62 @@ func deploymentName(podName string) string {
...
+	if err != nil {
+		logger.Error(err, "Could not get the service")
+		return ""
+	}
+	if svc.Spec.Type != corev1.ServiceTypeLoadBalancer {
+		logger.Error(errors.New("unexpected service type"), "Service type is not LoadBalancer")
+		return ""
+	}
+
+	externalAddrs := make([]string, 0, 4)
```
SeriyBg, line 207: Is there a reason why the default capacity is 4? Maybe it's better to set it to `len(svc.Status.LoadBalancer.Ingress)`?

SeriyBg (changes requested):
Added of few not so critical comments

controllers/util/util.go
hakanmemisoglu: I will update it to the maximum item possible. I was just trying to get a constant that will cover most of the cases.

controllers/util/predefined.go
```diff
@@ -0,0 +1,11 @@
+package util
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+)
+
+type Predefined interface {
```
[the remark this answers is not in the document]
hakanmemisoglu: Makes sense!

SeriyBg (approved):
LGTM

mtyazici (approved):
LGTM as well. I have one question, why didn't we also include the NodePort service's external IP addresses. We could get a node's IP address and show it.

hakanmemisoglu:
@mtyazici These are the conditions we agreed on before starting the task. I think we can add your idea as a next task but I don't think publicly exposing node ports is a common scenario.

PR #137, opened 2022-02-17
[CN-318 | CN-317] Removed publishing to marketplace and running e2e test during publishing process

cheels:
Since we decided we don't publish Operator to the marketplace, this part was removed from the CI process. Also, it's unnecessary to run e2e tests during the publishing process because it takes a lot of time, and if we need it, we have a separate job for it. 

no ci


mtyazici (approved):
LGTM

PR #141, opened 2022-02-21
E2E tests: remove storage class to use the default one

SeriyBg:
The tests contained the hardcoded name of `standard` ClassName which is the default one for GKE but is not present for EKS and AKS


mtyazici (approved):

hakanmemisoglu (approved):

hakanmemisoglu:
@SeriyBg Did you update with `make manifests` and `make generate`? 
I thought `+nullable` would change the CRD.

PR #149, opened 2022-03-03
[CN-273] Fix default CRD field values and comments

mtyazici:
Changes:
- Add enum tags from string enum types
- Use pointers for struct fields*
- Add omitempty for fields we don't want to use default go values
- Add default spec value for MC and HZ CRs to allow empty spec
- Remove applyDefaultValues functions from MC and HZ CRs 

\* This is needed because when empty structs with `omitempty` tag gets marshaled into json they don't get omitted but put as `{}`. Which in turn, activates the default values for the struct which we don't want because the user has not provided any input for the struct. This was present for `Persistence` field in Hazelcast CR as well as other struct fields.


controllers/naming/constants.go
```diff
@@ -57,9 +57,9 @@ const (
 	// DefaultClusterName default name of Hazelcast cluster
 	DefaultClusterName = "dev"
 	// HazelcastRepo image repository for Hazelcast
-	HazelcastRepo = "hazelcast/hazelcast"
```
hakanmemisoglu, line 60: Are these changes necessary? They have the same meaning as far as I know, and the older one is more concise.

hakanmemisoglu (approved):

controllers/naming/constants.go
mtyazici: There was a discrepancy between kubebuilder default values and default values in `constants.go`. I just chose one of them. We could also use the older one, I don't have a preference.

controllers/naming/constants.go
```diff
@@ -57,9 +57,9 @@ const (
...
+	HazelcastRepo = "docker.io/hazelcast/hazelcast"
 	// HazelcastEERepo image repository for Hazelcast EE
-	HazelcastEERepo = "hazelcast/hazelcast-enterprise"
+	HazelcastEERepo = "docker.io/hazelcast/hazelcast-enterprise"
 	// HazelcastVersion version of Hazelcast image
 	HazelcastVersion = "5.0.2"
```
hasancelik, line 64: All constants whose default values are configured by kubebuilder are redundant, I think. Changing the value of them does not mean anything because they are used by only tests files. Should we move them to test package? And how can we keep them in sync with kubebuilder values?

controllers/hazelcast/hazelcast_resources_test.go
```diff
@@ -4,104 +4,13 @@ import (
...
 )
 
-func Test_mergeHazelcastSpecs(t *testing.T) {
-	tests := []struct {
-		name   string
-		target hazelcastv1alpha1.HazelcastSpec
-		want   hazelcastv1alpha1.HazelcastSpec
-	}{
-		{
-			name:   "Empty hazelcast repository",
```
hasancelik, line 24: I think keeping these tests still is a good idea. WDYT?

controllers/managementcenter/managementcenter_resource_test.go
```diff
@@ -69,71 +63,3 @@ func Test_clusterAddCommand(t *testing.T) {
...
-		LicenseKeySecret: n.LicenseKeySecret,
-		Repository:       n.MCRepo,
-	}
-
-	tests := []struct {
-		name   string
-		target hazelcastv1alpha1.ManagementCenterSpec
-		want   hazelcastv1alpha1.ManagementCenterSpec
-	}{
-		{
```
hasancelik, line 85: same

hasancelik (changes requested):

api/v1alpha1/hazelcast_types.go
```diff
@@ -24,50 +25,54 @@ const (
 // HazelcastSpec defines the desired state of Hazelcast
 type HazelcastSpec struct {
 	// Number of Hazelcast members in the cluster.
-	// +kubebuilder:validation:Minimum=0
+	// +kubebuilder:validation:Minimum=1
```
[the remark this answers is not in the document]
mtyazici: Then, we need to make the clusterSize a pointer. I will change it

controllers/naming/constants.go
mtyazici: We have integration tests that ensure that our kubebuilder default values and default values in go files are the same. If not, our tests will fail and we will fix the values.

controllers/hazelcast/hazelcast_resources_test.go
mtyazici: Also we still have integration tests doing a very similar job to this test.

controllers/managementcenter/managementcenter_resource_test.go
mtyazici: Answered above.

hasancelik (approved):

PR #153, opened 2022-03-15
[CN-339] Fix HostPath E2E test for Hazelcast Persistence

mtyazici:
Changes
- Added privileges to pods in OCP environment when `hostPath` is enabled.
- Added a step to E2E tests to delete the`/tmp/hazelcast` folder in the host filesystem. 
- Refactored the `hostPath` E2E tests and added a new one for testing `hostPath` using multiple nodes.


test/e2e/util_test.go
```diff
@@ -69,3 +78,139 @@ func deleteIfExists(name types.NamespacedName, obj client.Object) {
...
+			Template: v1.PodTemplateSpec{
+				ObjectMeta: metav1.ObjectMeta{
+					Labels: lb,
+				},
+				Spec: v1.PodSpec{
+					Containers: []v1.Container{{
+						Name:    "cleanup",
+						Image:   "ubuntu",
+						Command: []string{"sh"},
+						Args:    []string{"-c", fmt.Sprintf("set -x; touch /host/%s ; sleep 5; rm -rf /host/%s2 || echo 'Could not delete'; echo 'done'; sleep 120", hzDir, hzDir)},
```
hakanmemisoglu, line 105: What is the purpose of `sleep 120`?
Is it related to the lazy reloading of files?

test/e2e/util_test.go
```diff
@@ -69,3 +78,139 @@ func deleteIfExists(name types.NamespacedName, obj client.Object) {
...
+func waitForDSPods(ds *appsv1.DaemonSet, lb client.MatchingLabels) *corev1.PodList {
+	pods := &corev1.PodList{}
+	podLabels := lb
+	Eventually(func() bool {
+		err := k8sClient.Get(context.Background(), types.NamespacedName{Name: ds.Name, Namespace: ds.Namespace}, ds)
+		if err != nil {
+			return false
+		}
+		return ds.Status.DesiredNumberScheduled != 0
+	}, timeout, interval).Should(BeTrue())
```
hakanmemisoglu, line 170: Is this block necessary?

mtyazici: Yes, that field does not get assigned immediately, we have to wait for DS controller to fill it with correct number, so in the next block we can compare it with created pod count.

test/e2e/util_test.go
mtyazici: I could not set DaemoSset pods' scheduling as `NeverRestart`, it is not permitted. So I put a sleep into containers after they are finished deleting the files. In the code we check the pod logs and when all of them are finished we delete the DS. The sleep time is quite arbitrary, probably 10 second sleep would also be enough.

hakanmemisoglu (approved):

hasancelik (approved):

PR #160, opened 2022-03-22
[CN-332] Do not delete the clusters if there is a failure

hakanmemisoglu:
Disable cleanup in nightlies if there is a failure


mtyazici:
Are we sure we want to do manual clean up for every failing nightly tests?

hakanmemisoglu:
It was the goal. The main goal is to provide a complete cluster state for debugging. I think it will also force us to come up with a fix as soon as possible. If we are overwhelmed by this change, we can revert it later on.

mtyazici (approved):
Alright, let's try it then :rocket:

hasancelik (approved):

PR #171, opened 2022-03-30
[CN-190] Hazelcast Members detailed status

SeriyBg:


test/e2e/hazelcast_controller_e2e_test.go
```diff
@@ -226,6 +226,29 @@ var _ = Describe("Hazelcast", func() {
 				evaluateReadyMembers(lookupKey, 3)
 			})
 		})
+
+		FIt("should update HZ detailed member status", func() {
```
dzeromski-hazelcast, line 230: "Fit" should be "It"? :)

go.mod
```diff
@@ -20,3 +19,5 @@ require (
 	k8s.io/utils v0.0.0-20210111153108-fddb29f9d009
 	sigs.k8s.io/controller-runtime v0.8.3
 )
+
+replace github.com/hazelcast/hazelcast-go-client v1.3.0 => github.com/yuce/hazelcast-go-client v1.3.0-preext3.0.20220324011025-02c0712e56c7
```
dzeromski-hazelcast, line 23: Should we use some private repo here?

SeriyBg: No. But for now, the changes that we are using aren't yet merged to the `hazelcast/hazelcast-go-client`. Thus the PR itself should be merged after the Go-client changes are released

Makefile
```diff
@@ -151,7 +151,7 @@ test-ph: generate fmt vet ginkgo ## Run phone-home tests
 
 ##@ Build
 
-GO_BUILD_TAGS ?= "localrun"
+GO_BUILD_TAGS ?= "hazelcastinternal localrun"
 build: generate fmt vet ## Build manager binary.
```
mtyazici, line 155: Could you also update build-tilt directive with the build tags?

test/e2e/hazelcast_controller_e2e_test.go
SeriyBg: Yeap, good catch! Changed it

Makefile
SeriyBg: Done! Also change the tag's settings, as `hazelcastinternal` should not be configurable

controllers/hazelcast/hazelcast_status_controller.go
```diff
@@ -59,34 +102,50 @@ func (c *HazelcastClient) start(ctx context.Context, config hazelcast.Config) {
...
+			c.MemberMap[changed.Member.UUID] = newMemberData(changed.Member)
 			c.Unlock()
 			c.Log.Info("Member is added", "member", changed.Member.String())
 		} else if changed.State == cluster.MembershipStateRemoved {
 			c.Lock()
-			delete(c.MemberMap, changed.Member.UUID.String())
+			delete(c.MemberMap, changed.Member.UUID)
 			c.Unlock()
 			c.Log.Info("Member is deleted", "member", changed.Member.String())
 		}
```
mtyazici, line 151: When I tested the status, I saw that the member statuses don't get enriched until the first ticker comes at 1 min after. Maybe we could add the updateMemberStates function here like the following:

```suggestion
		}
		  c.updateMemberStates(context.Background())
```

controllers/hazelcast/hazelcast_status_controller.go
```diff
@@ -101,3 +160,60 @@ func (c *HazelcastClient) triggerReconcile() {
...
+		if err != nil {
+			c.Log.Error(err, "TimedMemberState json parsing failed.", "CR", c.NamespacedName, "JSON", jsonState)
+			continue
+		}
+		m.enrichMemberData(state.TimedMemberState)
+	}
+	c.triggerReconcile()
+}
+
+func fetchTimedMemberState(ctx context.Context, client *hazelcast.Client, uuid hztypes.UUID) (string, error) {
```
mtyazici, line 185: Do you think we should move `fetchTimedMemberState` function and its related structs to a separate file? How should we continue when we add `addMapConfig` and other codecs?

mtyazici (approved):
LGTM, just left two comments

dzeromski-hazelcast (approved):

controllers/hazelcast/hazelcast_status_controller.go
```diff
@@ -101,3 +160,60 @@ func (c *HazelcastClient) triggerReconcile() {
 			Name:      c.NamespacedName.Name,
 		}}}
 }
+
+func (c *HazelcastClient) updateMemberStates(ctx context.Context) {
+	if c.client == nil {
+		return
+	}
+	c.Log.Info("Updating Hazelcast status", "CR", c.NamespacedName)
```
hasancelik, line 168: @SeriyBg Our status update logs are a little bit noisy due to ticker:

Screen Shot 2022-03-31 at 16 42 59

hasancelik (changes requested):

controllers/hazelcast/hazelcast_status_controller.go
SeriyBg: I'm not sure we should. The codecs are in a separate package, but the `fetchTimedMemberState` is related to the `status`. So I would not mix `fetchTimedMemberState` and `addMapConfig` into the same file but rather separate it by the logic. WDYT?

mtyazici: I did not say we should mix them in the same file but in separate files. Maybe we will not need it as you say. Better to keep it like you did.

controllers/hazelcast/hazelcast_status_controller.go
hasancelik: I realized the same thing, is there any reason behind that? @SeriyBg

hasancelik (approved):

controllers/hazelcast/hazelcast_status_controller.go
mtyazici: Yes, its because we only enrich the memberData in updateMemberStates function and it is only called in the ticker goroutine.

SeriyBg: Good catch, thanks @mtyazici! However, I've implemented it a bit differently, to enrich each member right after the new member event. With this, we will evitate the redundant call to already enriched members

PR #176, opened 2022-04-07
[CN-220] Improve operator status for client

mtyazici:
Changes:
- Fixed race condition between `initHzClient` and `shutdown`, where if `StartNewClientWithConfig` takes too much time reconcile function gets blocked for a very long time.
- Improved Hazelcast status messages about client's connection to the cluster. 

TODO:
- [ ] Operator still cannot understand when a connected client gets disconnected from the cluster. We need to use the functions in the client introduced in this commit https://github.com/hazelcast/hazelcast-go-client/commit/6845efd6ac3d6826c75e23ed75705b74c48a21aa.

### DIFF
--- a/controllers/hazelcast/hazelcast_controller.go
+++ b/controllers/hazelcast/hazelcast_controller.go
@@ -187,7 +187,9 @@ func (r *HazelcastReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 
 	externalAddrs := util.GetExternalAddresses(ctx, r.Client, h, logger)
-	return update(ctx, r.Client, h, r.runningPhaseWithMembers(req).withExternalAddresses(externalAddrs))
+	return update(ctx, r.Client, h, r.runningPhaseWithMembers(req).
+		withExternalAddresses(externalAddrs).
+		withMessage(clientConnectionMessage(req)))
 }
 
 func (r *HazelcastReconciler) runningPhaseWithMembers(req ctrl.Request) optionsBuilder {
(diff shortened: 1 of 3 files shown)


api/v1alpha1/zz_generated.deepcopy.go
```diff
@@ -1,3 +1,4 @@
+//go:build !ignore_autogenerated
```
hakanmemisoglu, line 1: I think you can delete this line.

hasancelik:
> TODO: [] Operator still cannot understand when a connected client gets disconnected from the cluster. We need to use the functions in the client introduced in this commit https://github.com/hazelcast/hazelcast-go-client/commit/6845efd6ac3d6826c75e23ed75705b74c48a21aa.

It is satisfied with https://github.com/hazelcast/hazelcast-platform-operator/pull/180, right?

mtyazici:
@hasancelik, I don't think it solves the problem. In that PR we do not check if the client is connected to the cluster but we only check member connection.

controllers/hazelcast/hazelcast_controller.go
```diff
@@ -226,6 +228,24 @@ func getHazelcastCRName(pod *corev1.Pod) (string, bool) {
 	}
 }
 
+func clientConnectionMessage(req ctrl.Request) string {
+	c, ok := GetClient(req.NamespacedName)
+	if !ok {
+		return "Operator failed to create connection to cluster . Some features might be unavailable."
```
hakanmemisoglu, line 234:
```suggestion
		return "Operator failed to create connection to cluster , some features might be unavailable."
```

hakanmemisoglu, line 234:
```suggestion
		return "Operator failed to create connection to cluster, some features might be unavailable."
```

hakanmemisoglu (approved):

SeriyBg (approved):

PR #177, opened 2022-04-07
[CN-319]Added FailFast mechanism

cheels:
- Added FailFast mechanism to unlabelled tests (fast test will be <= 100 sec and for slow is >=100 seconds accordingly)
- Marked tests with slow and fast labels


go.mod
```diff
@@ -9,7 +9,6 @@ require (
 	github.com/onsi/ginkgo/v2 v2.1.3
 	github.com/onsi/gomega v1.18.1
 	github.com/robfig/cron/v3 v3.0.0
-	github.com/stretchr/testify v1.6.1
```
mtyazici, line 12: Why did we have this one?

mtyazici (approved):

go.mod
cheels: I don't remember. Most probably to use some assertion from testify.

SeriyBg (approved):

PR #180, opened 2022-04-08
[CN-336] Add member list update functionality

cagric0:
- Used go client's HZ internal client to get member list and to check the connectivity of a member
- Go client added the newly functions to the master, so go.mod is updated
- e2e test is reverted to the original one and the tests were run in GKE

### DIFF
--- a/controllers/hazelcast/hazelcast_status_controller.go
+++ b/controllers/hazelcast/hazelcast_status_controller.go
@@ -109,6 +109,7 @@ func (c *HazelcastClient) start(ctx context.Context, config hazelcast.Config) {
 			case <-s.done:
 				return
 			case <-s.ticker.C:
+				c.updateMemberList()
 				c.updateMemberStates(ctx)
 			}
 		}
@@ -147,7 +148,6 @@ func (c *HazelcastClient) shutdown(ctx context.Context) {
 func getStatusUpdateListener(ctx context.Context, c *HazelcastClient) func(cluster.MembershipStateChanged) {
 	return func(changed cluster.MembershipStateChanged) {
 		if changed.State == cluster.MembershipStateAdded {
-			c.updateMemberList()
 			c.Lock()
 			m := newMemberData(changed.Member)
 			c.enrichMemberByUuid(ctx, changed.Member.UUID, m)
@@ -205,13 +205,13 @@ func (c *HazelcastClient) updateMemberList() {
 
 	memberList := hzInternalClient.OrderedMembers()
 
-	activeMembers := make(map[hztypes.UUID]int8, len(c.MemberMap))
+	activeMembers := make(map[hztypes.UUID]struct{}, len(c.MemberMap))
 
 	for _, memberInfo := range memberList {
 		_, ok := c.MemberMap[memberInfo.UUID]
 		if ok {
 			if hzInternalClient.ConnectedToMember(memberInfo.UUID) {
-				activeMembers[memberInfo.UUID] = 1
+				activeMembers[memberInfo.UUID] = struct{}{}
 			}
 		}
 	}
--- a/controllers/hazelcast/hazelcast_status_controller.go
+++ b/controllers/hazelcast/hazelcast_status_controller.go
@@ -17,7 +17,7 @@ import (
 	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/controller-runtime/pkg/event"
 
-	"github.com/hazelcast/hazelcast-platform-operator/api/v1alpha1"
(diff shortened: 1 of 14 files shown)


controllers/hazelcast/hazelcast_status_controller.go
```diff
@@ -143,6 +147,7 @@ func (c *HazelcastClient) shutdown(ctx context.Context) {
 func getStatusUpdateListener(ctx context.Context, c *HazelcastClient) func(cluster.MembershipStateChanged) {
 	return func(changed cluster.MembershipStateChanged) {
 		if changed.State == cluster.MembershipStateAdded {
+			c.updateMemberList()
```
SeriyBg, line 150: Just wondering if this is the best place to call the `updateMemberList`. As this will be called only for each `MembersipStateAdded` event. Maybe we can trigger this function periodically, probably in the same `ticker` that we fetch the member data (e.g. `updateMemberStates`)?

hakanmemisoglu: I agree with this. Actually, this one will not fix the problem we see right now, if the events are missed, we will not update the member list.

controllers/hazelcast/hazelcast_status_controller.go
```diff
@@ -192,6 +197,36 @@ func (c *HazelcastClient) enrichMemberByUuid(ctx context.Context, uuid hztypes.U
...
 
+func (c *HazelcastClient) updateMemberList() {
+	if c.client == nil {
+		return
+	}
+	hzInternalClient := hazelcast.NewClientInternal(c.client)
+
+	memberList := hzInternalClient.OrderedMembers()
+
+	activeMembers := make(map[hztypes.UUID]int8, len(c.MemberMap))
```
hakanmemisoglu, line 208: One trick for implementing set: https://emersion.fr/blog/2017/sets-in-go/
`make(map[hztypes.UUID],bool)` will be more explanatory. 
`make(map[hztypes.UUID],struct{})` will be a little bit more efficient.
It is up to you, you can keep this version too.

controllers/hazelcast/hazelcast_status_controller.go
cagric0: I was expecting these comments. I thought we may miss only `MembershipStateRemoved` events that's why I made this call in `MembershipStateAdded` event. Thus, I wanted to avoid extra calls in the ticker. But as you commented, it's more proper solution. I'll update the PR

controllers/hazelcast/hazelcast_status_controller.go
```diff
@@ -192,6 +197,35 @@ func (c *HazelcastClient) enrichMemberByUuid(ctx context.Context, uuid hztypes.U
...
+
+	activeMembers := make(map[hztypes.UUID]struct{}, len(c.MemberMap))
+
+	for _, memberInfo := range memberList {
+		_, ok := c.MemberMap[memberInfo.UUID]
+		if ok {
+			if hzInternalClient.ConnectedToMember(memberInfo.UUID) {
+				activeMembers[memberInfo.UUID] = struct{}{}
+			}
+		}
```
SeriyBg, line 216:
```suggestion
		if ok && hzInternalClient.ConnectedToMember(memberInfo.UUID) {
			activeMembers[memberInfo.UUID] = struct{}{}
		}
```

controllers/hazelcast/hazelcast_status_controller.go
```diff
@@ -192,6 +197,35 @@ func (c *HazelcastClient) enrichMemberByUuid(ctx context.Context, uuid hztypes.U
...
+	}
+	c.Lock()
+	for uuid, m := range c.MemberMap {
```
SeriyBg, line 219: Not sure this scenario will ever occur, but what if the `OrderedMembers()` will contain the member that is not present in the `c.MemberMap`? It seems like for now it won't add this member to it. I think we should. WDYT?

SeriyBg (approved):
Two small comments. Other than that LGTM

hakanmemisoglu (approved):
LGTM.
I agree on the comment by @SeriyBg about missed MemberAdded events. 
Maybe, you can add it in a followup PR.

controllers/hazelcast/hazelcast_status_controller.go
```diff
@@ -192,6 +197,35 @@ func (c *HazelcastClient) enrichMemberByUuid(ctx context.Context, uuid hztypes.U
...
+		_, ok := activeMembers[uuid]
+		if !ok {
+			delete(c.MemberMap, uuid)
```
mtyazici, line 222: Instead of deleting members, shouldn't we mark the member not reachable? The connection information client returns is only related to the client not the cluster itself, am I wrong?

PR #182, opened 2022-04-08
[CN-343]-Refactoring-e2e-tests

cheels:


test/e2e/helpers_test.go
```diff
@@ -0,0 +1,348 @@
...
+			seq = match[i]
+		}
+	}
+	if seq == "" {
+		Fail("Backup sequence not found")
+	}
+	return seq
+}
+
+func ReadLogs(scanner *bufio.Scanner, matcher types2.GomegaMatcher) {
```
SeriyBg, line 65: Do we really need this `func`? For me reading `EventyuallyInLogs().Should` is clearer than `ReadLogs()`

test/e2e/helpers_test.go
```diff
@@ -0,0 +1,348 @@
...
+
+	hazelcastcomv1alpha1 "github.com/hazelcast/hazelcast-platform-operator/api/v1alpha1"
+	n "github.com/hazelcast/hazelcast-platform-operator/controllers/naming"
+	"github.com/hazelcast/hazelcast-platform-operator/controllers/platform"
+	"github.com/hazelcast/hazelcast-platform-operator/test"
+)
+
+func GetBackupSequence(t time.Time) string {
+	By("Finding Backup sequence")
+	logs := test.GetPodLogs(context.Background(), types.NamespacedName{
```
SeriyBg, line 38: We can reuse `InitLogs` here, can't we?

SeriyBg (commented):
Left a few comments that are rather minor

test/e2e/helpers_test.go
cheels: Yes, good catch! 👍

test/e2e/helpers_test.go
cheels: I don't know, for me, ReadLogs looks more compact without digging into details about timeout that we nowhere configuring in the tests. But with upcoming changes with Eventually timeout EventyuallyInLogs() might be better and more flexible.
Revert this to initial state

test/e2e/helpers_test.go
```diff
@@ -0,0 +1,348 @@
...
+		log.Fatal(err)
+	}
+	clientSet, err := kubernetes.NewForConfig(restConfig)
+	if err != nil {
+		log.Fatal(err)
+	}
+	return clientSet
+}
+
+func ChangeHzClusterState(clusterState string, interval, timeout time.Duration) string {
```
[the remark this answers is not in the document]
cheels: Agree! I added ClusterState type

test/e2e/helpers_test.go
```diff
@@ -0,0 +1,343 @@
...
+
+	hzClient "github.com/hazelcast/hazelcast-go-client"
+
+	hazelcastcomv1alpha1 "github.com/hazelcast/hazelcast-platform-operator/api/v1alpha1"
+	n "github.com/hazelcast/hazelcast-platform-operator/controllers/naming"
+	"github.com/hazelcast/hazelcast-platform-operator/controllers/platform"
+	"github.com/hazelcast/hazelcast-platform-operator/test"
+)
+
+type ClusterState string
```
SeriyBg, line 35: We have the `ClusterState` type under `hazelcast`. Could we reuse it instead of redefining it?

PR #183, opened 2022-04-09
[CN-342]OLM-bundle-integration-pre-release test

cheels:
Added the next tests which is run separately in parallel:
Test type | Description
-- | --
kiwi | Full operator test
lemon | Full test of operator to be deployed from scratch
orange | Full test of operator to be deployed with existing bundles in quay registry


.github/workflows/redhat_operator_tests.yml
```diff
@@ -0,0 +1,100 @@
...
+    name: Build and Publish
+    defaults:
+      run:
+        shell: bash
+    runs-on: ubuntu-latest
+    steps:
+      - name: Validate version
+        run: |
+          NUMBER='(0|[1-9][0-9]*)'
+          echo ${RELEASE_VERSION} | egrep "^$NUMBER.$NUMBER(.$NUMBER)?$"
```
mtyazici, line 28: This check will fail when the workflow gets triggered by the cron job.

cheels: Yes, I noticed that and already fixed it in the latest commit

.github/workflows/redhat_operator_tests.yml
```diff
@@ -0,0 +1,100 @@
+name: Red Hat Operator Tests
+on:
+  workflow_dispatch:
+    inputs:
+      RELEASE_VERSION:
+        description: "Version in the form of X.Y[.Z]"
+        required: true
+        default: 7.8.9
```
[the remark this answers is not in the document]
cheels: I agree. Deleted it

PR #186, opened 2022-04-11
[CN-342]-OLM bundle integration pre release test

cheels:


.github/workflows/redhat_operator_tests.yml
```diff
@@ -72,6 +75,7 @@ jobs:
       OPP_DEBUG: 1
       OPP_PRODUCTION_TYPE: k8s
       RELEASE_VERSION: ${{ needs.publish_image.outputs.RELEASE_VERSION }}
```
mtyazici, line 77: I think we are not using the RELEASE_VERSION anymore, maybe we can remove it.

mtyazici: Yeah, but you should  also remove the related parts from the first job.

cheels: Removed from everywhere

mtyazici (approved):

PR #190, opened 2022-04-14
Increase the timeout to fix timeout problems in OCP and AKS

hakanmemisoglu:
Increase the overall and the eventually block timeouts to fix problem in OCP and AKS


cagric0 (commented):
LGTM. One reminder question. Should we create an issue for detecting the execution time for each test to customize eventually timeout?

hakanmemisoglu:
@cagric0 I will open one.

PR #191, opened 2022-04-15
BR agent sidecar support

hasancelik:


dzeromski-hazelcast (approved):

controllers/hazelcast/br_agent_rest_client.go
```diff
@@ -0,0 +1,87 @@
...
+	for _, address := range ac.addresses {
+		req, err := postRequestWithBody(ctxT, reqBody, address, uploadBackup)
+		if err != nil {
+			return fmt.Errorf("request creation failed: %s, address --> %q , URL --> %q ", err, address, address+uploadBackup)
+		}
+		res, err := ac.executeRequest(req)
+		if err != nil {
+			return err
+		}
+		defer res.Body.Close()
```
cagric0, line 63: Do we need to call `Close` function? I think we don't need since we don't open(read) `res.Body`.

cagric0 (approved):
I have two little question. Other than that LGTM!

api/v1alpha1/hotbackup_types.go
```diff
@@ -38,7 +38,11 @@ type HotBackupSpec struct {
 	//	@daily (or @midnight)  | Run once a day, midnight                   | 0 0 * * *
 	//	@hourly                | Run once an hour, beginning of hour        | 0 * * * *
 	// +optional
-	Schedule string `json:"schedule,omitempty"`
+	Schedule string `json:"schedule"`
+
+	// URL of the bucket to download HotBackup folders.
+	// +optional
+	BucketURL string `json:"bucket"`
```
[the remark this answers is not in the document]
hasancelik: Users can create HotBackup CR without external storage

controllers/hazelcast/br_agent_rest_client.go
hasancelik: You do not need to open the body, it is used by the request itself.

PR #193, opened 2022-04-15
Makefile updates for Go 1.18 support and Apple silicon

hasancelik:
- upgrade kustomize version to 4.5.3
- use 'go install' instead of 'go get' because go get no longer builds or installs packages in module-aware mode(https://tip.golang.org/doc/go1.18#go-command)

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -233,7 +233,7 @@ controller-gen: ## Download controller-gen locally if necessary.
 
 KUSTOMIZE = $(shell pwd)/bin/kustomize
 kustomize: ## Download kustomize locally if necessary.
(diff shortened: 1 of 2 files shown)


hakanmemisoglu:
@hasancelik Should we update go version everywhere? (Dockerfiles etc.)

hakanmemisoglu:
@hasancelik Should we update go version everywhere? (Dockerfiles etc.)

hasancelik:
> @hasancelik Should we update go version everywhere? (Dockerfiles etc.)

@hakanmemisoglu I think these tasks should be handled by another task/PR(we need to investigate operator SDK compatibility etc). The changes at this PR are just local setup-related changes.

hakanmemisoglu (approved):

mtyazici (approved):

PR #195, opened 2022-04-18
[CN-288] Add map CR

mtyazici:
Changes:
- Added codecs for `addMapConfig`, `getMapConfig` and `updateMapConfig`.
- Added separate `types` folder for types used in codec functions.
- Added Map CR and controller reconcile logic.
- Added E2E and integration tests for the Map CR.

TODO:
- [ ] Add persistence of the map config when map is created/updated successfully for each member. It will be implemented in a different PR.


api/v1alpha1/map_types.go
```diff
@@ -0,0 +1,214 @@
...
+	SchemeBuilder.Register(&Map{}, &MapList{})
+}
+
+var EncodeInMemoryFormat map[types.InMemoryFormat]int32 = map[types.InMemoryFormat]int32{
+	types.InMemoryFormatBinary: 0,
+	types.InMemoryFormatObject: 1,
+	types.InMemoryFormatNative: 2,
+}
+
+var EncodeMaxSizePolicy map[types.MaxSizePolicyType]int32 = map[types.MaxSizePolicyType]int32{
```
hakanmemisoglu, line 184:
```suggestion
var EncodeMaxSizePolicy = map[types.MaxSizePolicyType]int32{
```
I think this should work. There are other lines in this file that you can apply it.

api/v1alpha1/map_types.go
```diff
@@ -0,0 +1,214 @@
...
+	// Name of the map config to be created. If empty, CR name will be used.
+	// It cannot be updated after map config is created successfully.
+	// +optional
+	Name string `json:"name,omitempty"`
+
+	// Count of synchronous backups.
+	// It cannot be updated after map config is created successfully.
+	// +kubebuilder:default:=1
+	// +optional
+	BackupCount *int32 `json:"backupCount,omitempty"`
```
hakanmemisoglu, line 20: What is the reason that these fields are pointers?

config/crd/patches/cainjection_in_maps.yaml
```diff
@@ -0,0 +1,7 @@
+# The following patch adds a directive for certmanager to inject CA into the CRD
```
hakanmemisoglu, line 1: We can remove the file. I don't think it is required (I could not find webhook implementation).

hakanmemisoglu: Also the other file in the directory.

controllers/naming/constants.go
```diff
@@ -76,6 +76,16 @@ const (
 	MCImagePullPolicy = corev1.PullIfNotPresent
 )
 
+// Map CR default values
+const (
+	MapBackupCount        = int32(1)
```
hakanmemisoglu, line 81: I think prefixing these constants with `Default` would be better.

config/crd/patches/cainjection_in_maps.yaml
mtyazici: I will remove it and the other file

controllers/naming/constants.go
mtyazici: Makes sense, I will update these constants.

api/v1alpha1/map_types.go
mtyazici: I wasn't aware this worked. I will also put all these fields into a `var` block.

api/v1alpha1/map_types.go
hakanmemisoglu: Makes sense. Thanks for the explanation.

api/v1alpha1/map_types.go
```diff
@@ -0,0 +1,214 @@
+package v1alpha1
+
+import (
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	"github.com/hazelcast/hazelcast-platform-operator/controllers/protocol/types"
+)
+
+// MapSpec defines the desired state of Hazelcast Map Config
+type MapSpec struct {
```
SeriyBg, line 10: Just out of curiosity, is there a reason why we decided to have the same config structure as the standard Hazelcast config and not to hide the complexity as we're doing for the other features?

SeriyBg (commented):
Posted one, rather conceptual, question

api/v1alpha1/map_types.go
mtyazici: We implement our features keeping the Hazelcast Clloud's offerings in mind. All of the fields here except `ReadBackupData` is configurable in the Hazelcast Cloud. I think exposing these fields are mandatory. However, I am open to exposing them in a more friendly way if you have any suggestions.

SeriyBg: I see. The cloud requirement should be kept in mind. But I was just confused with all the config options; as for the new users, it would be something very complex, which we are trying to avoid. Not that I have a clear suggestion on how to avoid it here, though 😄

mtyazici: Yeah,  it is hard to avoid complexity here :D I think it comes down to providing good default values and promoting only the necessary fields in the Map CR documentation page. These fields will be documented in the API Reference page but I don't think we need to promote every one of them in the main page.

config/crd/patches/cainjection_in_maps.yaml
mtyazici: I checked it, we have mentioned files for Hazelcast and MC. What do you think we should do?

hakanmemisoglu: This is not very important. It is up to you.

controllers/util/util.go
```diff
@@ -237,3 +238,46 @@ func getLoadBalancerAddress(lb *corev1.LoadBalancerIngress) string {
 	}
 	return ""
 }
+
+func IndexConfigSliceEquals(a, b []hazelcastv1alpha1.IndexConfig) bool {
```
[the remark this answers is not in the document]
mtyazici: I am not sure if I understood, could you elaborate?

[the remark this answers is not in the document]
hakanmemisoglu: For example, `a == [c1, c2],  b == [c2, c1]`

controllers/hazelcast/map_controller.go
```diff
@@ -0,0 +1,259 @@
...
+			continue
+		}
+		memberStatuses[member.UUID.String()] = hazelcastv1alpha1.MapSuccess
+	}
+	errString := failedMembers.String()
+	if errString == "" {
+		return memberStatuses, nil
+	}
+
+	return memberStatuses, fmt.Errorf("Error creating/updating the Map config %s for members %s.", m.MapName(), errString[:len(errString)-2])
```
SeriyBg, line 195:
```suggestion
	if errString != "" {
		return memberStatuses, fmt.Errorf("Error creating/updating the Map config %s for members %s.", m.MapName(), errString[:len(errString)-2])
	}

	return memberStatuses, nil
```

This is very minor suggestion just to follow [the happy path to the left edge](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88) idiomatic rule

SeriyBg (approved):
Apart from two small comments, LGTM

hakanmemisoglu (approved):

api/v1alpha1/map_types.go
```diff
@@ -0,0 +1,222 @@
...
+	// +kubebuilder:default:={maxSize: 0}
+	// +optional
+	Eviction *EvictionConfig `json:"eviction,omitempty"`
+
+	// Used to enable reading from local backup map entries.
+	// It can be used if there is at least 1 sync or async backup.
+	// It can be updated.
+	// +kubebuilder:default:=false
+	// +optional
+	ReadBackupData bool `json:"readBackupData"`
```
hasancelik, line 54: Why this one is configurable? just because it can be updated with `updateMapConfig` method as you mentioned at the TDD?

api/v1alpha1/map_types.go
```diff
@@ -0,0 +1,222 @@
...
+
+	// Data type that will be used for storing the map data.
+	// +kubebuilder:default:=BINARY
+	// +optional
+	InMemoryFormat types.InMemoryFormat `json:"inMemoryFormat,omitempty"`
```
hasancelik, line 59: We do not have `NativeMemoryConfig` option at Hazelcast CR so `NATIVE` option is useless at this point, right?
https://docs.hazelcast.com/hazelcast/5.1/data-structures/setting-data-format#using-high-density-memory-store-with-map

api/v1alpha1/map_types.go
```diff
@@ -0,0 +1,222 @@
...
+	// +kubebuilder:default:=1
+	// +optional
+	BackupCount *int32 `json:"backupCount,omitempty"`
+
+	// Number of asynchronous backups. Unlike the synchronous backup process,
+	// asynchronous backup process does not block the map operations.
+	// It cannot be updated after map config is created successfully.
+	// +kubebuilder:default:=0
+	// +optional
+	AsyncBackupCount *int32 `json:"asyncBackupCount,omitempty"`
```
hasancelik, line 27: As far as I can understand, AsyncBackupCount is not supported by Cloud. Should we keep it?

api/v1alpha1/map_types.go
mtyazici: I reverted the inMemoryFormat changes.

api/v1alpha1/map_types.go
mtyazici: I removed the `readBackupData` from the CR.

api/v1alpha1/map_types.go
```diff
@@ -0,0 +1,204 @@
...
+	// You can learn more at https://docs.hazelcast.com/hazelcast/latest/query/indexing-maps.
+	// It cannot be updated after map config is created successfully.
+	// +optional
+	Indexes []IndexConfig `json:"indexes,omitempty"`
+
+	// When enabled, map data will be persisted.
+	// It cannot be updated after map config is created successfully.
+	// +kubebuilder:default:=false
+	// +optional
+	PersistenceEnabled bool `json:"persistenceEnabled"`
```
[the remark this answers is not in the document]
mtyazici: I added code for persistence validation.

hasancelik (approved):

PR #196, opened 2022-04-19
[CN-306] Define restore agent config types

dzeromski-hazelcast:


api/v1alpha1/hazelcast_types.go
```diff
@@ -96,6 +101,23 @@ type BackupAgentConfiguration struct {
...
+type RestoreAgentConfiguration struct {
+	// Repository to pull Hazelcast Platform Operator Agent(https://github.com/hazelcast/platform-operator-agent)
+	// +kubebuilder:default:="docker.io/hazelcast/platform-operator-agent"
+	// +optional
+	AgentRepository string `json:"agentRepository,omitempty"`
+
+	// Version of Hazelcast Platform Operator Agent.
+	// +kubebuilder:default:="1.0.0"
+	// +optional
+	AgentVersion string `json:"agentVersion,omitempty"`
```
cagric0, line 114: Backup and Restore agents are the same agent. Why don't we use single _AgentRepository_ and _AgentVersion_?

api/v1alpha1/hazelcast_types.go
```diff
@@ -77,6 +77,11 @@ type HazelcastSpec struct {
 	// +optional
 	//+kubebuilder:default:={}
 	Backup BackupAgentConfiguration `json:"backup,omitempty"`
+
+	// Restore Agent configuration
+	// +optional
+	//+kubebuilder:default:={}
+	Restore RestoreAgentConfiguration `json:"restore,omitempty"`
```
hasancelik, line 84: It should be pointer, see https://github.com/hazelcast/hazelcast-platform-operator/pull/195#discussion_r853264539 and 
https://github.com/hazelcast/hazelcast-platform-operator/blob/28c1d27efb9195b35533ba5dd8951cbb2660d934/api/v1alpha1/hazelcast_types.go#L79
After pointer change, IsEnabled() should be also updated, see https://github.com/hazelcast/hazelcast-platform-operator/blob/28c1d27efb9195b35533ba5dd8951cbb2660d934/api/v1alpha1/hazelcast_types.go#L304

hasancelik (changes requested):

api/v1alpha1/hazelcast_types.go
dzeromski-hazelcast: This is the requirement from TDD. I guess there could be a case where no backup is defined and only restore is.

config/samples/_v1alpha1_hazelcast_persistence_restore.yaml
```diff
@@ -0,0 +1,21 @@
...
+  licenseKeySecret: hazelcast-license-key
+  persistence:
+    baseDir: "/data/hot-restart/"
+    clusterDataRecoveryPolicy: "FullRecoveryOnly"
+    pvc:
+      accessModes: ["ReadWriteOnce"]
+      requestStorage: 8Gi
+      storageClassName: "standard"
+  restore:
+    agentRepository: dzeromskhazelcast/platform-operator-agent
```
dzeromski-hazelcast, line 18: Will need to update this before final commit. Right now we point to latest development agent.

config/samples/_v1alpha1_hazelcast_persistence_restore.yaml
```diff
@@ -0,0 +1,21 @@
...
+    agentVersion: latest
+    bucketSecret: br-secret
+    bucketPath: "s3://hazelcast-cn-306-restore-tests?prefix=hazelcast/2022-02-21-14-23-52/&region=us-east-1"
```
dzeromski-hazelcast, line 21: Again, need to change it before final version. This is just an example.

api/v1alpha1/hazelcast_types.go
dzeromski-hazelcast: Fixed

hasancelik (changes requested):

cagric0 (approved):
LGTM!

controllers/hazelcast/hazelcast_resources.go
```diff
@@ -704,6 +709,61 @@ func backupAgentContainer(h *hazelcastv1alpha1.Hazelcast) v1.Container {
...
+					SecretKeyRef: &v1.SecretKeySelector{
+						LocalObjectReference: v1.LocalObjectReference{
+							Name: h.Spec.Restore.BucketSecret,
+						},
+						Key: n.BucketDataS3SecretAccessKey,
+					},
+				},
+			},
+			{
+				Name: "AWS_SESSION_TOKEN",
```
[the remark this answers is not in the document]
dzeromski-hazelcast: It was due to SSO requirement. Will delete this as in productions it wont be used. Good catch.

api/v1alpha1/hazelcast_types.go
```diff
@@ -101,6 +106,27 @@ type BackupAgentConfiguration struct {
 	BucketSecret string `json:"bucketSecret,omitempty"`
```
cagric0, line 106: Same with restore. This also shouldn't be optional I think.

api/v1alpha1/hazelcast_types.go
```diff
@@ -101,6 +106,27 @@ type BackupAgentConfiguration struct {
...
+	// +optional
+	AgentVersion string `json:"agentVersion,omitempty"`
+
+	// Name of the secret with credentials for cloud providers.
+	// +optional
+	BucketSecret string `json:"bucketSecret,omitempty"`
+
+	// Full path to blob storage bucket.
+	// +optional
+	BucketPath string `json:"bucketPath,omitempty"`
```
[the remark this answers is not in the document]
dzeromski-hazelcast: We can discuss this with @hasancelik.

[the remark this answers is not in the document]
cagric0: We already talked actually but we can have a call together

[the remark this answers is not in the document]
hasancelik: As @cagric0 mentioned, we talked about it, and making them non-optional is more reasonable for UX.

hasancelik (changes requested):

api/v1alpha1/hazelcast_types.go
dzeromski-hazelcast: Fixed.

api/v1alpha1/hazelcast_types.go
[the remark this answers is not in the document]
dzeromski-hazelcast: Fixed.

hasancelik (approved):

controllers/hazelcast/hazelcast_resources.go
```diff
@@ -798,6 +837,43 @@ func (r *HazelcastReconciler) checkHotRestart(ctx context.Context, h *hazelcastv
...
+
+	// check if all cluster members are in passive state
+	for _, member := range h.Status.Members {
+		if ClusterState(member.State) != Passive {
+			return nil
+		}
+	}
+
+	rest := NewRestClient(h)
+	state, err := rest.GetState(ctx)
```
SeriyBg, line 867: Just wondering, could we just check the `h.Status.Members` instead of making the REST call?

controllers/hazelcast/hazelcast_resources.go
```diff
@@ -798,6 +837,43 @@ func (r *HazelcastReconciler) checkHotRestart(ctx context.Context, h *hazelcastv
...
+	if err != nil {
+		return err
+	}
+	if state != "passive" {
```
SeriyBg, line 871: What if the state is `frozen` or any other state apart from `passive`? Is this the acceptable scenario for us?

controllers/hazelcast/hazelcast_resources.go
```diff
@@ -798,6 +837,43 @@ func (r *HazelcastReconciler) checkHotRestart(ctx context.Context, h *hazelcastv
...
+		return nil
+	}
+
+	if h.Status.Phase == hazelcastv1alpha1.Pending {
+		return nil
+	}
+
+	// check if all cluster members are in passive state
+	for _, member := range h.Status.Members {
+		if ClusterState(member.State) != Passive {
```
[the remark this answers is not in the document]
SeriyBg: To add to this. In the `TimedMemberState` we have a field `"clusterState": "ACTIVE"`. Maybe we could just add the `clusterState` to the `HazelcastStatus` and update in the `updateMembers` func in the `hazelcast_status_controller`? This will simplify this whole method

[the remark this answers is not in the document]
dzeromski-hazelcast: We could but this is new to me. Can we do the simplification and adding new state property in a new PR?

controllers/hazelcast/hazelcast_resources.go
dzeromski-hazelcast: We are checking members state few lines before. This rest call is just to avoid stale state before we try to activate.

controllers/hazelcast/hazelcast_resources.go
dzeromski-hazelcast: This is undefined for our case so we ignore it. Proper restore will always start nodes in passive state, cluster in passive state with restore status set and pending phase. If anything differs it basically failed restore for us and we do nothing then.

PR #197, opened 2022-04-19
[CN-306] Add restore initContainer spec

dzeromski-hazelcast:


cagric0 (approved):
LGTM! 
But I have questions about _LivenessProbe_ and _ReadinessProbe_ (line 640, 654).
What are they and why do we have them for only backup not for restore.
My idea for what they are, is kubernetes makes api call for given endpoints within the thresholds and periods basically health check right?

dzeromski-hazelcast:
Replaced with #197

PR #200, opened 2022-04-25
Fix Slack notify job triggers

mtyazici:
Current triggers for Slack jobs in some of the workflows, only get triggered from the `delete-cluster` job. This was OK, when we were not deleting the clusters if an E2E test failed. However, currently `delete-cluster` jobs always run and it will not trigger the Slack notify job even though the E2E tests fail.

### DIFF
--- a/.github/workflows/nightly-e2e-aks.yaml
+++ b/.github/workflows/nightly-e2e-aks.yaml
@@ -114,10 +114,10 @@ jobs:
         run: |
           make clean-up-namespace NAMESPACE=${NAMESPACE}
 
-  clean-up:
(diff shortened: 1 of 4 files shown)


hasancelik:
@mtyazici so we will receive notification for green builds also, right?

mtyazici:
No, only failed workflows will trigger the Slack notification job.

hasancelik (commented):
@mtyazici so we will receive notification for green builds also, right?

hasancelik (approved):

hakanmemisoglu (approved):

PR #201, opened 2022-04-26
Remove CRD creationg from E2E test code

mtyazici:
E2E tests in the CI sometimes fail when applying the CRDs from the code because we are running `EE` and `OS` tests at the same time in the same cluster. 

CRD creation step in the code is not necessary since we are doing it before running the tests.

An example test failure is here https://github.com/hazelcast/hazelcast-platform-operator/runs/6190621996?check_suite_focus=true

### DIFF
--- a/test/e2e/e2e_suite_test.go
+++ b/test/e2e/e2e_suite_test.go
@@ -32,9 +32,7 @@ var _ = BeforeSuite(func() {
 	logf.SetLogger(zap.New(zap.WriteTo(GinkgoWriter), zap.UseDevMode(true)))
 
 	By("bootstrapping test environment")
-	testEnv = &envtest.Environment{
-		ErrorIfCRDPathMissing: false,
-	}
+	testEnv = &envtest.Environment{}
 
 	cfg, err := testEnv.Start()
 	Expect(err).NotTo(HaveOccurred())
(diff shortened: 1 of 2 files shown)


test/e2e/e2e_suite_test.go
```diff
@@ -34,8 +33,7 @@ var _ = BeforeSuite(func() {
 
 	By("bootstrapping test environment")
 	testEnv = &envtest.Environment{
-		CRDDirectoryPaths:     []string{filepath.Join("../..", "config", "crd", "bases")},
-		ErrorIfCRDPathMissing: true,
+		ErrorIfCRDPathMissing: false,
```
SeriyBg, line 36: Isn't `false` the default value here?

SeriyBg (commented):
Won't we need to add the same change for Phone Home test suite?

test/e2e/e2e_suite_test.go
mtyazici: You are right, I guess we don't need to give it. Somehow, I wanted it to be explicit :D

mtyazici:
>  Won't we need to add the same change for Phone Home test suite?

You are right, I forgot about that one. I added it also.

SeriyBg (approved):
LGTM

cagric0 (approved):
LGTM!

PR #202, opened 2022-04-27
Refactor struct name to apply go conventions

hakanmemisoglu:
This PR changes the struct name from `HazelcastClient` to `Client` since the package name is already `hazelcast`. 

Additionally, the locking in `initHzClient` function is moved to the line after we create the internal Hazelcast client.


controllers/hazelcast/hazelcast_status_controller.go
```diff
@@ -149,10 +148,10 @@ func (c *HazelcastClient) start(ctx context.Context, config hazelcast.Config) {
 	}(ctx, c.statusTicker)
 }
 
-func (c *HazelcastClient) initHzClient(ctx context.Context, config hazelcast.Config) {
+func (c *Client) initHzClient(ctx context.Context, config hazelcast.Config) {
+	hzClient, err := hazelcast.StartNewClientWithConfig(ctx, config)
```
hakanmemisoglu, line 152: @mtyazici What was the exact problem here we discussed before?

mtyazici: We were running `StartNewClientWithConfig` after we locked. If this function took too much time and user decided to delete the Hazelcast CR, operator could not delete the client because the the other goroutine could not take the lock.

mtyazici: In the deletion part I had changed the order of `cancel` function and locking, so the issue I mentioned was fixed. This change also removes that issue,

hakanmemisoglu: OK, thanks for the reminder.

controllers/hazelcast/hazelcast_status_controller.go
```diff
@@ -94,28 +94,26 @@ func CreateClient(ctx context.Context, h *hazelcastv1alpha1.Hazelcast, channel c
 	}
 	config := buildConfig(h)
 	c := newHazelcastClient(l, ns, channel)
-	config.AddMembershipListener(getStatusUpdateListener(ctx, c))
```
SeriyBg, line 97: I'm a bit confused, why are we removing this listener?

PR #205, opened 2022-05-03
[CN-359] Set HotBackup status to Pending after creation and before the start

SeriyBg:


controllers/hazelcast/hot_backup_controller.go
```diff
@@ -268,8 +276,18 @@ func (r *HotBackupReconciler) executeFinalizer(ctx context.Context, hb *hazelcas
...
+func (r *HotBackupReconciler) triggerHotBackup(ctx context.Context, req reconcile.Request, rest *RestClient, logger logr.Logger) error {
+	hb := &hazelcastv1alpha1.HotBackup{}
+	err := r.Get(ctx, req.NamespacedName, hb)
+	if err != nil {
+		return err
+	}
+	if !hb.Status.State.IsRunning() {
+		hb.Status.State = hazelcastv1alpha1.HotBackupPending
+		_ = r.Status().Update(ctx, hb)
+	}
```
mtyazici, line 288: Why are we not using the existing hot backup object from the reconciler function?

SeriyBg: The reason is that the `HotBackup` object might be updated by the status update function, thus we need to fetch the last version of it before the update

mtyazici: I see, this only happens when HotBackup is scheduled with a cron job, right? Also, what happens when user removes the cron job from the CR? From what I get, the cron job will continue and HotBackup also will get triggered once for the change.

dzeromski-hazelcast (approved):

controllers/hazelcast/hot_backup_controller_test.go
```diff
@@ -136,13 +137,80 @@ func TestHotBackupReconciler_shouldSetStatusToFailedWhenHbCallFails(t *testing.T
...
+
+	Eventually(func() hazelcastv1alpha1.HotBackupState {
+		_ = r.Client.Get(context.TODO(), n, hb)
+		return hb.Status.State
+	}, 2*time.Second, 100*time.Millisecond).Should(Equal(hazelcastv1alpha1.HotBackupPending))
+
+	reconcileWg.Add(1)
+	go func() {
+		defer reconcileWg.Done()
+		_, _ = r.Reconcile(context.TODO(), reconcile.Request{NamespacedName: n})
```
dzeromski-hazelcast, line 206: nit: skip the assignment?

SeriyBg: We can skip it, but Goland told me not to do so 😃 
image

mtyazici (approved):

PR #207, opened 2022-05-04
[CN-341]Added unified pipeline to release in Docker and Red Hat registries

cheels:
- Added unified pipeline to release in Docker and Red Hat registries
- Added numbers for pipelines to get a more clear vision of what should be run at first
- Renamed workflow file docker hub-release to publish release (now it contains all release steps)
- Added composite action that allows to set up all required tools
- Added pre-release tests using preflight and community tests (kiwi, orange, lemon)
- Updated base UBI image in Docker file from 8.5 to 8.6 tag to get the health index 'A' from RedHat
- Added OpenSCAP scan for the test image before release


.github/scripts/actions/operator-tests/action.yml
```diff
@@ -0,0 +1,91 @@
...
+    - name: Install operator SDK
+      shell: bash
+      run: |
+        if [[ ! -z "${{ inputs.OPERATOR_SDK_VERSION }}" ]]; then
+          echo "✅ Installing OPERATOR_SDK"
+          OPERATOR_SDK=${GITHUB_WORKSPACE}/bin/operator-sdk
+          curl -sSL https://github.com/operator-framework/operator-sdk/releases/download/${{ inputs.OPERATOR_SDK_VERSION }}/operator-sdk_linux_amd64 -o $OPERATOR_SDK --create-dirs && chmod +x $OPERATOR_SDK
+        else
+           echo "⏭ Skipping OPERATOR_SDK installation"
+        fi
```
mtyazici, line 91: We can install `operator-sdk` using `Makefile` as well, just like opm.

.github/workflows/operator-certification-tests.yaml
```diff
@@ -0,0 +1,201 @@
...
+jobs:
+  test_container:
+    name: Build and Test Container
+    defaults:
+      run:
+        shell: bash
+    runs-on: ubuntu-latest
+    outputs:
+      RELEASE_VERSION: ${{ steps.setup-envs.outputs.RELEASE_VERSION }}
+      CONTAINER_REPOSITORY: ${{ steps.setup-envs.outputs.CONTAINER_REPOSITORY }}
```
mtyazici, line 25: There is no output set up in `setup-envs` as `CONTAINER_REPOSITORY`.

.github/workflows/operator-certification-tests.yaml
```diff
@@ -0,0 +1,201 @@
...
+            docker login -u "${{ secrets.DOCKERHUB_USERNAME }}" --password-stdin
+
+      - name: Login to OCP and Create New Project
+        run: |
+          oc login ${OCP_CLUSTER_URL} -u=${OCP_USERNAME} -p=${OCP_PASSWORD} --insecure-skip-tls-verify
+          oc new-project ${NAMESPACE}
+
+      - name: Initiating an Operator Bundle Scan
+        run: |
+          preflight check operator ${BUNDLE_IMAGE}
```
mtyazici, line 167: Does preflight tool needs an Openshift environment to test the bundle image?

.github/workflows/community-operator-tests.yaml
```diff
@@ -0,0 +1,46 @@
+name: (3) Community Operator Tests
```
mtyazici, line 1: We have a very similar workflow to this one, maybe we can merge them into one workflow with multiple triggers. We can have if/else statements for the different parts. WDYT?

mtyazici (commented):
@cheels, thanks for the work. I left some questions regarding to the workflows, LGTM otherwise :+1:

.github/scripts/actions/operator-tests/action.yml
cheels: The last time when I did it it didn't work for me. But now it works, so will fix it

.github/workflows/operator-certification-tests.yaml
cheels: Thx, removed

.github/workflows/community-operator-tests.yaml
cheels: I thought we can remove that (old) workflow completely. I think the better is to make new workflow [operator-certification-tests.yaml] more flexible  to run the tests only by a cron job and using workflow_dispatch

.github/workflows/operator-certification-tests.yaml
```diff
@@ -0,0 +1,268 @@
...
+          grep -E -q "Preflight result: PASSED\"( |$)" preflight.log || exit 1
+
+      - name: Clean Up Openshift Namespace
+        run: |
+          oc delete project ${NAMESPACE}
```
mtyazici, line 229: This command will wait for the deletion of the namespace. I am not sure but if there is any problem this can block the workflow.

.github/workflows/operator-certification-tests.yaml
```diff
@@ -0,0 +1,268 @@
...
+    with:
+      BUNDLE_VERSION: ${{ needs.build_bundle.outputs.BUNDLE_VERSION }}
+
+  release_operator:
+    if: github.event_name == 'push' && startsWith(github.ref, 'refs/tags/v')
+    needs: [test_container, test_operator, community_operators_test_suite]
+    uses: ./.github/workflows/publish-release.yaml
+    with:
+      RELEASE_VERSION: ${{ needs.test_container.outputs.RELEASE_VERSION }}
+      PROJECT_ID: ${{ needs.test_container.outputs.PROJECT_ID }}
```
mtyazici, line 243: Just to double check, the reason we are getting `PROJECT_ID` and `PREFLIGHT_VERSION` from the needed job outputs instead of using `env` is that `env` context is not available within `jobs.<job-id>.with`, right?

.github/workflows/operator-certification-tests.yaml
```diff
@@ -0,0 +1,268 @@
...
+  push:
+    tags:
+      - "v*"
+  workflow_dispatch:
+    inputs:
+      RELEASE_VERSION:
+        description: "Version in the form of X.Y[.Z]"
+        required: true
+  schedule:
+    - cron: "0 9 * * *"
```
mtyazici, line 12: Do you think daily schedule is not an overkill?

mtyazici (commented):
LGTM overall :+1: thanks for the great work. Though our release process just got a lot more complex :D.

I have one question left though. In `operator-certification-tests.yaml`, we are also releasing the operator if the trigger is a tag push. I think the workflow name is a bit misleading. We can maybe change the name of the workflow or separate the releasing part into another workflow, wdyt?

.github/workflows/operator-certification-tests.yaml
```diff
@@ -83,9 +83,50 @@ jobs:
...
+    defaults:
+      run:
+        shell: bash
+    env:
+      CONTAINER_IMAGE: ${{ needs.test_container.outputs.CONTAINER_IMAGE }}
+    runs-on: ubuntu-latest
+    steps:
+      - name: Run OpenSCAP Vulnerability Scan
+        id: vulnerabilityScan
+        uses: appleboy/ssh-action@master
```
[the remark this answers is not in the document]
cheels: Oh, nice catch. I just forgot to change it to v0.1.4

.github/workflows/operator-certification-tests.yaml
cheels: Yes exactly, this is a necessary measure, unfortunately. GitHub Actions doesn't allow at the moment to pass the variables from global env into **Reusing Workflows** **workflow_call**. For example the next won't work:

```
env:
  PROJECT_ID: ${{ secrets.PROJECT_ID }}

release_operator:
    uses: ./.github/workflows/publish-release.yaml
    with:
      PROJECT_ID: ${PROJECT_ID}
```

.github/workflows/operator-certification-tests.yaml
cheels: It was not my decision honestly😁 What is your suggestion? Once per 2 days?

mtyazici: I learned the tests take 14 mins in total, I think its fine then to have it daily.

hasancelik (approved):
LGTM 👍 Thanks for your detailed work @cheels 👏 🙂

PR #208, opened 2022-05-04
[CN-343]-Refactoring-e2e-tests

cheels:
- Created helpers_test.go that contains all required functions.
- The test class now contains only tests that increase readability


test/e2e/helpers_test.go
```diff
@@ -0,0 +1,323 @@
...
+)
+
+func GetBackupSequence(t time.Time) string {
+	By("Finding Backup sequence")
+	logs := InitLogs(t)
+	scanner := bufio.NewScanner(logs)
+	test.EventuallyInLogs(scanner, timeout, logInterval).Should(ContainSubstring("Starting new hot backup with sequence"))
+	line := scanner.Text()
+	defer logs.Close()
+	Expect(logs.Close()).Should(Succeed())
```
SeriyBg, line 44: We are calling `logs.Close()` twice, and the one in the `Expect` block will be called immediately. Is it expected?

test/e2e/helpers_test.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+	Expect(addr).Should(Not(BeEmpty()))
+	By("connecting Hazelcast client")
+	config := hzClient.Config{}
+	config.Cluster.Network.SetAddresses(fmt.Sprintf("%s:5701", addr))
+	config.Cluster.Unisocket = unisocket
+	config.Cluster.Discovery.UsePublicIP = true
+	client, err := hzClient.StartNewClientWithConfig(ctx, config)
+	Expect(err).ToNot(HaveOccurred())
+	return client, err
```
SeriyBg, line 137: We can only return the `client` as `err` will always be `nil`, if it's not `nil` the `func` will fail one line above

test/e2e/helpers_test.go
```diff
@@ -0,0 +1,323 @@
...
+	if err != nil {
+		log.Fatal(err)
+	}
+	return clientSet
+}
+
+func FillTheMapData(ctx context.Context, unisocket bool, mapName string, mapSize int) {
+	var m *hzClient.Map
+	clientHz, err := GetHzClient(ctx, unisocket)
+	Expect(err).ToNot(HaveOccurred())
```
SeriyBg, line 156: This is redundant as we have `Expect(err).ToNot(HaveOccurred())` inside the `GetHzClient` function

test/e2e/helpers_test.go
```diff
@@ -0,0 +1,323 @@
...
+	client, err := hzClient.StartNewClientWithConfig(ctx, config)
+	Expect(err).ToNot(HaveOccurred())
+	return client, err
+}
+
+func GetClientSet() *kubernetes.Clientset {
+	kubeConfig := clientcmd.NewNonInteractiveDeferredLoadingClientConfig(clientcmd.NewDefaultClientConfigLoadingRules(), &clientcmd.ConfigOverrides{})
+	restConfig, err := kubeConfig.ClientConfig()
+	if err != nil {
+		log.Fatal(err)
```
mtyazici, line 144: I assume Failing or Expecting nil here instead of logging would be better. WDYT?

test/e2e/helpers_test.go
```diff
@@ -0,0 +1,323 @@
...
+	}
+	clientSet, err := kubernetes.NewForConfig(restConfig)
+	if err != nil {
+		log.Fatal(err)
```
mtyazici, line 148: Same as above.

test/e2e/helpers_test.go
mtyazici: I also think parametrizing `GetHzClient` and `FillTheMapData` for Hazelcast CR NamespacedName would be beneficial for filling different Hazelcast clusters.

test/e2e/helpers_test.go
SeriyBg: `log.Fatal` is actually failing the proccess

mtyazici: Ah, didn't know that. :+1:

SeriyBg: Yeap, that's a bit counterintuitive. I only know this as we already discussed it with @cheels once 😃

mtyazici: Got it :D

test/e2e/helpers_test.go
cheels: Fixed

test/e2e/helpers_test.go
cheels: @SeriyBg I removed it, thx! 
@mtyazici Decided to leave it as it is for the moment

test/e2e/helpers_test.go
cheels: Fixed

SeriyBg (approved):

PR #209, opened 2022-05-09
[CN-374] Improve client to cluster connection check

mtyazici:


controllers/hazelcast/hazelcast_status.go
```diff
@@ -127,7 +127,7 @@ func update(ctx context.Context, c client.Client, h *hazelcastv1alpha1.Hazelcast
 
 	cl, ok := GetClient(types.NamespacedName{Name: h.Name, Namespace: h.Namespace})
 
-	if ok && cl.client != nil && cl.client.Running() {
+	if ok && cl.client != nil && util.IsClientConnected(hazelcast.NewClientInternal(cl.client)) {
```
hakanmemisoglu, line 130: This check would be easier to read. 
```suggestion
	if ok && cl.IsClientConnected() {
```

mtyazici: You are right, I will change it to that.

hakanmemisoglu (approved):

PR #214, opened 2022-05-11
[CN-335] Use different zap logger options for local and prod envs

dzeromski-hazelcast:


mtyazici (commented):
I think we should not assume `hazelcast_config.go` is only used for production. When operator is deployed in the cluster it will use `hazelcast_config.go` and I am deploying the operator for developing and testing.

SeriyBg (commented):
In addition to @mtyazici's comment, maybe it make sense to use the option flags instead of the programmatic approach?

SeriyBg (approved):

mtyazici (approved):
Just left a trivial comment.

PR #216, opened 2022-05-11
[CN-368] Enable WAN reference in Map config

hakanmemisoglu:
Enable WAN reference in Map configuration by pointing to the default WAN configuration


mtyazici (approved):
LGTM :+1:, maybe we can add wan-replication-ref check into one of the e2e tests for map

controllers/hazelcast/map_controller.go
```diff
@@ -256,7 +262,33 @@ func fillAddMapConfigInput(mapInput *codecTypes.AddMapConfigInput, m *hazelcastv
 	}
 	mapInput.IndexConfigs = copyIndexes(ms.Indexes)
 	mapInput.HotRestartConfig.Enabled = ms.PersistenceEnabled
+	mapInput.WanReplicationRef = defaultWanReplicationRefCodec(hz, m)
+}
+
+func defaultWanReplicationRefCodec(hz *hazelcastv1alpha1.Hazelcast, m *hazelcastv1alpha1.Map) codecTypes.WanReplicationRef {
+	if !util.IsEnterprise(hz.Spec.Repository) {
```
SeriyBg, line 269: Wouldn't it be better to fail the configuration if the cluster is not EE?

SeriyBg (approved):
LGTM

controllers/hazelcast/map_controller.go
hakanmemisoglu: @SeriyBg, wan-reference is a part under map configuration and we need to fill it under every condition because it is not an updatable field. Also, this field is not exposed to the user, so the user has no control over this field. 

There will be wan-configuration which wan-reference points to in the next PRs. It will be tied to the new WanConfiguration API. In this case, I am planning to implement the check you mentioned. If the Hazelcast cluster is not EE, and the user creates a WanConfiguration for a Map in the non-EE HZ cluster, it will fail quickly.

cagric0 (approved):
LGTM!

PR #221, opened 2022-05-16
[CN-373] Separate E2E test suite into multiple suites

mtyazici:
Changes:
- Separated the test suite into multiple files. Every file is now a suite. There are following suites: `hz`, `mc`, `hz_persistence`, `hz_expose_externally`, `map`, `map_persistence`.
- Suites are defined with labels. You can run single or multiple suites with following commands. `make test-e2e E2E_TEST_SUITE=hz || mc`. To run all test suites just run `make test-e2e`.
- I added `-p` option to Ginkgo CLI. This will find the core count of the host and create as many processes to parallelize the tests. This somehow limits the parallelization if the core count is limited. This is the case for our GitHub actions hosted runners, which have 2 cores. 

I tried to run all test suites in matrices in the workflow. However, I got unexpected errors. If we want to improve the parallelization, we should look more into using matrices in the workflows.


test/e2e/hazelcast_exposeExternally_test.go
```diff
@@ -0,0 +1,86 @@
+package e2e
```
SeriyBg, line 1: Is there a reason why the file name is a mix of snake-case and camel-case?

test/e2e/hazelcast_test.go
```diff
@@ -0,0 +1,125 @@
...
+
+	hazelcastcomv1alpha1 "github.com/hazelcast/hazelcast-platform-operator/api/v1alpha1"
+	hazelcastconfig "github.com/hazelcast/hazelcast-platform-operator/test/e2e/config/hazelcast"
+)
+
+const (
+	logInterval = 10 * time.Millisecond
+)
+
+var _ = FDescribe("Hazelcast", Label("hz"), func() {
```
SeriyBg, line 23: Here as well should be `Describe`?

test/e2e/hazelcast_exposeExternally_test.go
mtyazici: Not particularly, I will update it. Thanks!

cagric0 (approved):
Thanks for the good job. LGTM!

SeriyBg (approved):
Other than a small nit, LGTM

PR #222, opened 2022-05-18
[CN-398] Set HB to failed when Hazelcast CR does not exist

SeriyBg:


hasancelik (approved):
LGTM 👍 We should add a HotBackup's message into the status also at scope of [Add message output in All CR statuses when output format is "-o wide"](https://hazelcast.atlassian.net/jira/software/projects/CN/boards/42/backlog?selectedIssue=CN-380) ticket.

dzeromski-hazelcast (approved):

PR #227, opened 2022-05-30
[CN-395] Make resources and scheduling spec fields updatable for HZ and MC

mtyazici:
With the current changes, `persistence` is the only field that updates are not reflected in the operator.


SeriyBg (approved):
LGTM, however, it would be great if we can have some tests for it (probably integration)

hakanmemisoglu (approved):
LGTM
+1 for integration test

mtyazici:
Sure, I will add integration tests for these.

PR #229, opened 2022-05-30
[CN-356, CN-319 ]-Separate timeout values for each Eventually block in E2E tests

cheels:
* Updated Eventually timeout for each block
* Added **-ginkgo.slow-spec-threshold=100s** to emit [SLOW] warning on tests if they exceed a certain threshold (default: 5 seconds).
* Added backup tests


Makefile
```diff
@@ -147,9 +147,9 @@ test-it: manifests generate fmt vet ## Run tests.
 	source ${ENVTEST_ASSETS_DIR}/setup-envtest.sh; fetch_envtest_tools $(ENVTEST_ASSETS_DIR); setup_envtest_env $(ENVTEST_ASSETS_DIR); PHONE_HOME_ENABLED=$(PHONE_HOME_ENABLED) DEVELOPER_MODE_ENABLED=$(DEVELOPER_MODE_ENABLED) go test -tags $(GO_BUILD_TAGS) -v ./test/integration/... -ginkgo.label-filter="slow || fast" -coverprofile cover.out $(GO_TEST_FLAGS) -timeout 5m
 
 test-e2e: generate fmt vet ginkgo ## Run end-to-end tests
-	USE_EXISTING_CLUSTER=true NAME_PREFIX=$(NAME_PREFIX) $(GINKGO) --tags $(GO_BUILD_TAGS) --vv --progress --timeout 60m --coverprofile cover.out ./test/e2e -- -ginkgo.label-filter="slow || fast" -namespace "$(NAMESPACE)" -eventually-timeout 15m  -delete-timeout 8m $(GO_TEST_FLAGS)
+	USE_EXISTING_CLUSTER=true NAME_PREFIX=$(NAME_PREFIX) $(GINKGO) --tags $(GO_BUILD_TAGS) --vv --progress --timeout 70m --coverprofile cover.out ./test/e2e -- -ginkgo.label-filter="slow || fast" -ginkgo.slow-spec-threshold=100s -namespace "$(NAMESPACE)" -eventually-timeout 15m  -delete-timeout 8m $(GO_TEST_FLAGS)
```
hakanmemisoglu, line 150:
```suggestion
	USE_EXISTING_CLUSTER=true NAME_PREFIX=$(NAME_PREFIX) $(GINKGO) --label-filter="slow || fast" --slow-spec-threshold=100s --tags $(GO_BUILD_TAGS) --vv --progress --timeout 70m --coverprofile cover.out ./test/e2e --  - -namespace "$(NAMESPACE)" -eventually-timeout 15m  -delete-timeout 8m $(GO_TEST_FLAGS)
```
According to Gingko docs, it might work too.

go.mod
```diff
@@ -20,4 +20,4 @@ require (
 )
 
 // to fix vulnerability: CVE-2021-3121 in github.com/gogo/protobuf < v1.3.2
-replace github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2
+replace github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2
```
hakanmemisoglu, line 23:
```suggestion
replace github.com/gogo/protobuf => github.com/gogo/protobuf v1.3.2

```

test/e2e/backup_test.go
```diff
@@ -0,0 +1,250 @@
...
+	"sigs.k8s.io/controller-runtime/pkg/client"
+
+	hzClient "github.com/hazelcast/hazelcast-go-client"
+
+	hazelcastcomv1alpha1 "github.com/hazelcast/hazelcast-platform-operator/api/v1alpha1"
+	"github.com/hazelcast/hazelcast-platform-operator/test"
+	hazelcastconfig "github.com/hazelcast/hazelcast-platform-operator/test/e2e/config/hazelcast"
+)
+
+var _ = Describe("Hazelcast", func() {
```
hakanmemisoglu, line 26:
```suggestion
var _ = Describe("Hazelcast Backup", func() {
```

test/e2e/hazelcast_controller_e2e_test.go
```diff
@@ -4,7 +4,7 @@ import (
 	"bufio"
 	"context"
 	"fmt"
-	"time"
+	. "time"
```
hakanmemisoglu, line 7: Why are we changing this one?

test/loadClient/Dockerfile
```diff
@@ -0,0 +1,10 @@
+FROM golang:1.17
+
+WORKDIR /app
+
+COPY go.mod ./
+COPY go.sum ./
+
+COPY *.go ./
+
+RUN go build -o /fill_map
```
hakanmemisoglu, line 10:
```suggestion
RUN go build -o /fill_map

```

test/loadClient/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/hazelcast/hazelcast-platform-operator/loadClient
+
+go 1.16
+
+require (
+	github.com/hazelcast/hazelcast-go-client v1.2.0
+	k8s.io/apimachinery v0.20.2
+	k8s.io/client-go v0.20.2
+)
```
hakanmemisoglu, line 9:
```suggestion
)

```

hakanmemisoglu (commented):
@cheels Just couple of questions. 
I will do a detailed review soon.

test/e2e/hazelcast_controller_e2e_test.go
cheels: I think it's more readable and doesn't require to write always the **time** word. 
Everybody knows that Seconds, Millisecond, and Minutes are always coming from the time package. 
So if you think is not necessary and it's better to always specify the time package like this I can revert this change. 

**Before**:
```
10 * time.Millisecond
```

**After**
```
10 * Millisecond
```

Makefile
cheels: Yes, it works also. Pushed the changes

test/loadClient/go.mod
```diff
@@ -0,0 +1,9 @@
+module github.com/hazelcast/hazelcast-platform-operator/loadClient
+
+go 1.16
```
hakanmemisoglu, line 3:
```suggestion
go 1.17
```
since the dockerfile use golang:1.17

hakanmemisoglu (approved):
LGTM, just a small comment

test/e2e/helpers_test.go
```diff
@@ -162,9 +165,62 @@ func FillTheMapData(ctx context.Context, unisocket bool, mapName string, mapSize
...
+	mapSize, _ := strconv.ParseFloat(mapSizeInGb, 64)
+	client := GetHzClient(ctx, false)
+	m, _ = client.GetMap(ctx, mapName)
+	Eventually(func() (int, error) {
+		return m.Size(ctx)
+	}, 15*Minute, interval).Should(Equal(int(math.Round(mapSize*1310.72) * 100)))
+	// 1310.72 entries per one Go routine. Formula: 1073741824 Bytes per 1Gb  / 8192 Bytes per entry / 100 go routines
+	err := client.Shutdown(ctx)
+	Expect(err).ToNot(HaveOccurred())
+	DeletePod(clientPod.Name, 0)
```
mtyazici, line 183: Maybe, deferring DeletePod is better in case of any errors.

cheels: As for a separate e2e suite. Yes, it is, backup_tests in different suite file but currently running in every e2e suite runs. 
Not sure I know how to run it separately without changing the workflow...

mtyazici: I think you are supposed to change the workflow, or use labels to exlude this suite from others.

mtyazici: > Why? It's a part of the function. The client pod will be always up and running and should be removed at the end of the flow. Shortly saying pod should always be removed in case of another test will use func FillTheMapWithHugeData() again and if pod still available the next test will be failed.

I mean it is already deleted at the end of the flow. However, if there is a problem in the flow, it will not be deleted. Should we not defer deleting it after we create it, if the pod's lifecycle ends in that function?

cheels: 1.  As for the defer call of the function it makes sense. I've made the change. Thanks!
2.  As for the running: we will agree to create a separate workflow and run it once per 3 days

mtyazici (approved):

PR #230, opened 2022-05-31
CN 313 Update B&R agent

cagric0:
In addition to S3, GS and Azure are also supported for the backup and restore.


config/samples/backup.yaml
```diff
@@ -0,0 +1,15 @@
+apiVersion: hazelcast.com/v1alpha1
```
hasancelik, line 1: Could you rename this file as `_v1alpha1_hotbackup_agent.yaml` then remove exsiting one?

controllers/hazelcast/hot_backup_controller.go
```diff
@@ -135,6 +135,10 @@ func (r *HotBackupReconciler) Reconcile(ctx context.Context, req reconcile.Reque
 	}
 
 	if hb.Spec.BucketURL != "" {
+		if hb.Spec.SecretName == "" {
```
hasancelik, line 138: This logic should be moved to [validation.go](https://github.com/hazelcast/hazelcast-platform-operator/blob/main/controllers/hazelcast/validation/validation.go)

config/samples/backup.yaml
cagric0: Is it okay to keep commented buckets and secrets?

dzeromski-hazelcast (approved):

config/samples/backup.yaml
hasancelik: I think so 👍

PR #231, opened 2022-06-02
CN-415 Move B&R configuration under persistence

cagric0:
Refactor `is backup and restore enable` logic.


dzeromski-hazelcast (approved):

controllers/hazelcast/hazelcast_resources.go
```diff
@@ -825,7 +827,7 @@ func restoreAgentCredentials(secret string, provider string) []v1.EnvVar {
 func backupAgentContainer(h *hazelcastv1alpha1.Hazelcast) v1.Container {
 	return v1.Container{
 		Name:  n.BackupAgent,
-		Image: h.BackupAgentDockerImage(),
+		Image: h.AgentDockerImage(),
```
mtyazici, line 830: Is there a possibility for this to panic because of nil reference?

test/integration/hazelcast_controller_test.go
```diff
@@ -796,10 +771,11 @@ var _ = Describe("Hazelcast controller", func() {
 						StorageClassName: &[]string{"standard"}[0],
 					},
 				}
-				spec.Backup = &hazelcastv1alpha1.BackupAgentConfiguration{
-					AgentRepository: "hazelcast/platform-operator-agent",
-					AgentVersion:    "0.1.0",
-					BucketSecret:    "br-secret",
+				// TODO: If we don't pass agent configuration it's not using default values
```
mtyazici, line 774: I think we might just give a default value for the agent configuration at the spec level if we have to do workaround like these, wdyt?

mtyazici (commented):
Added two comments about the agent configuration, other than that LGTM.

PR #232, opened 2022-06-03
CN420-create the separate workflow to run extended backup tests

cheels:
* Created a separate workflow to run the slow backup tests in the morning at 9 A.M once per 3 days. 
* Increase the number of the map size up to 10Gb to test the backup and restore


dzeromski-hazelcast (approved):

cagric0 (approved):
LGTM! We already talked but just as a reminder. We need to add fast tests for both local and external backup on nightly tests. If we are not planning to add them now, we may create a jira issue for it.

PR #244, opened 2022-06-15
[CN-350] Add init container for downloading jars into classpath

mtyazici:


dzeromski-hazelcast (approved):

internal/naming/constants.go
```diff
@@ -72,6 +73,8 @@ const (
 	GCP   = "gs"
 	AWS   = "s3"
 	AZURE = "azblob"
+
+	UserCustomClassPath = "/opt/hazelcast/customClass"
```
dzeromski-hazelcast, line 77: The only thing that I don't like is that we mix the names. Sometimes we use CustomClass and sometimes UserCustomClass.

controllers/hazelcast/hazelcast_resources.go
```diff
@@ -728,6 +728,17 @@ func (r *HazelcastReconciler) reconcileStatefulset(ctx context.Context, h *hazel
...
 		}
+
+		if h.Spec.CustomClass.IsEnabled() {
+			sts.Spec.Template.Spec.Volumes = append(sts.Spec.Template.Spec.Volumes, customClassVolume(h))
+			provider, err := hazelcastv1alpha1.BucketConfiguration(*h.Spec.CustomClass).GetProvider()
+			if err != nil {
+				return err
+			}
+			sts.Spec.Template.Spec.InitContainers = append(sts.Spec.Template.Spec.InitContainers, ccdAgentContainer(h, provider))
+			sts.Spec.Template.Spec.Containers[0].VolumeMounts = append(sts.Spec.Template.Spec.Containers[0].VolumeMounts, ccdAgentVolumeMount(h))
```
dzeromski-hazelcast, line 739: Wouldn't it be easier if we always had volume mount in hazelcast container instead of injecting this if custom class is enabled? It will be easier to reason about and will keep everything in one place.

dzeromski-hazelcast, line 739: Wouldn't it be easier if we always had volume mount in hazelcast container instead of injecting this if custom class is enabled? It will be easier to reason about and will keep everything in one place.

mtyazici: I am not quite sure if it is easier to reason about, could you provide an example? My thoughts were because it is only used for custom class feature, we should only mount it when it is enabled. Do you think it makes it more complicated this way?

mtyazici: I ended up changing it to your suggestion

api/v1alpha1/hazelcast_types.go
```diff
@@ -82,9 +82,22 @@ type HazelcastSpec struct {
...
 	Agent *AgentConfiguration `json:"agent,omitempty"`
+
+	// Custom Classes to Download into Class Path
+	// +optional
+	CustomClass *CustomClassConfiguration `json:"customClass,omitempty"`
+}
+
+type BucketConfiguration struct {
+	// Name of the secret with credentials for cloud providers.
+	Secret string `json:"secret,omitempty"`
```
hakanmemisoglu, line 93: You can add kubebuilder validations to both fields if you want.

mtyazici: To make sure they are not empty, right?

hakanmemisoglu: Yes, you can also use other ones if they applies to the requirements.

hakanmemisoglu (approved):
LGTM. I have only one small question. 
What happens both restore and ccd is enabled?
Do you handle the behavior of the agent with env variables?

mtyazici:
> What happens both restore and ccd is enabled?
> Do you handle the behavior of the agent with env variables?

If both of them are enabled, there will be two init containers running. But, we still use different environment variables so if we change the agent command's structures, we could run them in the same init container too.

api/v1alpha1/hazelcast_types.go
mtyazici: Thanks for the suggestion, I also removed validation code for this struct and put it into kubebuilder.

PR #249, opened 2022-06-20
CN-376-release-operator-v5.3

cheels:
* Added function to check image health index before publishing on RedHat 
* Added function to remove the published image (docker.io) and revert the previous image tag to the latest 
* Added function to remove release tag (Github) in case of failure 
* Added function to remove published bundle from Jfrog in case of failure and revert the previous bundle to latest 
* Added dependency to skip creating PR in case of the image was not published to RedHat 
* Added '--fail' option to all curl calls


mtyazici (approved):
LGTM :+1:

cheels:
Closing this one because of duplicate https://github.com/hazelcast/hazelcast-platform-operator/pull/250

PR #252, opened 2022-06-21
[CN-433] Wrap check into eventually block

hakanmemisoglu:
The check needs to be in eventually block since we are missing updated config for some test runs.


mtyazici (approved):
Solution looks cool, thanks for the fix :)

cagric0 (approved):

PR #256, opened 2022-06-23
Limit E2E test parallel process count to 4 and add makefile directive for focused tests

mtyazici:


hakanmemisoglu:
Can you use a variable with default value 4?

hakanmemisoglu:
Meanwhile, can you check whether we still use `eventually-timeout` and if not, can you remove it?

mtyazici:
Sure, I will do it

hakanmemisoglu (approved):

PR #269, opened 2022-06-30
Added concurrency to pull-request.yml workflow

cheels:
* Added cancellation option to workflow which makes it possible to cancel the workflow run if one more push was in the same PR
* Removed 'linter' dependency from 'unit-tests' job. Now, 'linter' and 'unit-tests' run in parallel.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -17,6 +17,10 @@ env:
   GCP_PROJECT_ID: hazelcast-33
   GKE_ZONE: europe-west1-b
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   linter:
     name: Run linters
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request.yml
```diff
@@ -265,14 +268,14 @@ jobs:
...
-        if: always()
+        if: ${{ !cancelled() && always() }}
         run: |
           make clean-up-namespace NAMESPACE=${NAMESPACE}
 
   delete-cluster:
     name: Delete Cluster
     runs-on: ubuntu-20.04
-    if: always() && needs.create-gke-cluster.result != 'skipped'
+    if: (!cancelled() && always() && needs.create-gke-cluster.result != 'skipped')
```
mtyazici, line 278: I think this `delete-cluster` should run always regardless if the workflow is cancelled.

.github/workflows/pull-request.yml
```diff
@@ -265,14 +268,14 @@ jobs:
           make test-e2e GO_TEST_FLAGS=${GO_TEST_FLAGS} NAMESPACE=$NAMESPACE NAME_PREFIX=$NAME_PREFIX
 
       - name: Clean up after Tests
-        if: always()
+        if: ${{ !cancelled() && always() }}
```
mtyazici, line 271: I think clean up step should run always regardless if the workflow is cancelled.

mtyazici (commented):
Thanks for the PR Alex. I have a question about the if conditions though. Are we certain that we need `${{ !cancelled() && always() }}` instead of just `${{ !cancelled() }}`?

.github/workflows/pull-request.yml
cheels: I performed few tests more and yes, you are right: in case of using 'if' and 'cancelled()' functions the 'always()' can be removed

dzeromski-hazelcast (approved):

mtyazici (approved):

PR #285, opened 2022-07-11
Updated workflows to run in parallel

cheels:


.github/workflows/backup-e2e-gke.yaml
```diff
@@ -37,26 +35,44 @@ jobs:
             --num-nodes=2
           sleep 30
 
+  get-image:
+    name: Get Image
```
dzeromski-hazelcast, line 39: nit: better to name it "build image"?

.github/workflows/backup-e2e-gke.yaml
```diff
@@ -37,26 +35,44 @@ jobs:
             --num-nodes=2
           sleep 30
 
+  get-image:
```
dzeromski-hazelcast, line 38: Why do we run it after linters? Can we do it in parallel to unit tests and linters?

.github/workflows/backup-e2e-gke.yaml
```diff
@@ -11,18 +11,16 @@ env:
 jobs:
   create-gke-cluster:
```
dzeromski-hazelcast, line 12: Can we run in in parallel to linters?

.github/workflows/pull-request.yml
```diff
@@ -274,7 +249,7 @@ jobs:
 
   delete-cluster:
     name: Delete Cluster
```
dzeromski-hazelcast, line 251: If we delete cluster can we skip step: "Clean up after Tests". Why bother with cleanup if in the end we delete everything?

.github/workflows/backup-e2e-gke.yaml
```diff
@@ -65,37 +81,16 @@ jobs:
...
-          IMG=ttl.sh/$(uuidgen):1h
-          echo "IMG=${IMG}" >> $GITHUB_ENV
-
-          make docker-build-ci IMG=$IMG VERSION=${{github.sha}}
-          make docker-push IMG=$IMG
+            --zone ${{ env.GKE_ZONE }} \
+            --project ${{ env.GCP_PROJECT_ID }}
 
       - name: Deploy Operator to GKE
         run: |
```
dzeromski-hazelcast, line 96: Can we increase ginkgo --procs 2? Would it improve test time?

.github/workflows/backup-e2e-gke.yaml
SeriyBg: Do we run the linters for the nightly tests? There is no point in it, as all the code must pass the PR builder, which does run the linters

.github/workflows/backup-e2e-gke.yaml
cheels: We don't have any linters in this workflow. What linters are you talking about?

.github/workflows/backup-e2e-gke.yaml
cheels: Guys, we don't have a linter step in nightly tests.

.github/workflows/backup-e2e-gke.yaml
cheels: As far as I remember, the 'ginkgo --procs 2' was changed by @mtyazici and was chosen for some performance reason. The more significant number can produce some performance issues. At least this param should be tested before change

.github/workflows/backup-e2e-gke.yaml
cheels: I changed the name from 'build image' to 'get image' because this job is not always 'build' the image but just returns the latest-snapshot name. So do you still think it's a good idea to name it "build-image"?

.github/workflows/backup-e2e-gke.yaml
```diff
@@ -37,26 +35,44 @@ jobs:
...
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Build and push image to ttl.sh
+        id: build-img
+        run: |
+          if [[ "${{github.event_name}}" == "workflow_dispatch" ]]; then
+            IMG=ttl.sh/$(uuidgen):2h
+            echo "IMG=${IMG}" >> $GITHUB_ENV
```
SeriyBg, line 52: It seems like adding the `IMG` to the env is redundant as this it the last step in the job, so it won't be used anywhere

.github/workflows/backup-e2e-gke.yaml
```diff
@@ -37,26 +35,44 @@ jobs:
...
+            echo "::set-output name=IMG::${IMG}"
+            make docker-build-ci IMG=$IMG VERSION=${{github.sha}}
+            make docker-push IMG=$IMG
+          else
+            IMG=hazelcast/hazelcast-platform-operator:latest-snapshot
+            echo "IMG=${IMG}" >> $GITHUB_ENV
```
SeriyBg, line 58: Same here

.github/workflows/pull-request.yml
cheels: That's a good point. "Clean up after Tests" probably could remove altogether from the workflow. Let me check.

.github/workflows/backup-e2e-gke.yaml
cheels: @dzeromski-hazelcast Or if you are talking about pull-request workflow and this dependency    
 needs: ['unit-tests','linter'] then I think it's required because if linters or unit tests are failed then there is no need to create a cluster and run the e2e test. Also, there are some conditions presented here:

      && needs.unit-tests.result == 'success'
      && needs.linter.result == 'success')

SeriyBg: @cheels well, it's better to run E2E tests even if linter fails. It's better to report all the errors at once, otherwise if the PR brakes some logic, the author won't know it before he fixes the lint errors

.github/workflows/backup-e2e-gke.yaml
cheels: Fixed

.github/workflows/backup-e2e-gke.yaml
SeriyBg: @cheels it is true if we want to save the execution time. But for the user experience, we won't report all the errors at once.
There is one more solution, we can split linter and E2E tests into different workflows. This way, the linter failures will be reported as soon as it finishes, while E2E tests will continue to run. Also in the UI it will shown as two separate failures, so the author of the PR can know why it failed even without checking the logs (although he will need to check the logs to know what to fix)

.github/workflows/backup-e2e-gke.yaml
dzeromski-hazelcast: Wasn't that limited because of problems on local machine? Can we have different value in CI?

.github/workflows/backup-e2e-gke.yaml
dzeromski-hazelcast: Well, yea. I was thinking about pull-request workflow but must have misclicked.

.github/workflows/backup-e2e-gke.yaml
cheels: Made changes to run linter, unit test, and image creation in parallel.

.github/workflows/backup-e2e-gke.yaml
cheels: Pushed updates

.github/workflows/backup-e2e-gke.yaml
cheels: So I tested with 6 procs. And the result was not successful: I got a lot of new failures with tests that are not failed before. I think this parameter should be tested a bit more not in this PR.

dzeromski-hazelcast: Can you create a task for it so we do not forget it?

cheels: @dzeromski-hazelcast Done. The task has been created.

PR #289, opened 2022-07-14
Disable lowDataMode as not needed

SeriyBg:
As for NewRelic settings, I have disabled scraping the needed GKE clusters data, which was the main reason for the huge amount of data ingestion. Without this, the `lowDataMode` is not needed as the data ingestion is less than 1GB a day for all the clouds


cheels (approved):
LGTM</br>_// approved in Slack with WorkerB_

PR #291, opened 2022-07-19
[CN-410] Add MapStore/MapLoader support for Map CR

mtyazici:


controllers/hazelcast/map_controller.go
```diff
@@ -245,7 +248,7 @@ func (r *MapReconciler) ReconcileMapConfig(
 	return memberStatuses, nil
 }
 
-func fillAddMapConfigInput(mapInput *codecTypes.AddMapConfigInput, hz *hazelcastv1alpha1.Hazelcast, m *hazelcastv1alpha1.Map) {
+func fillAddMapConfigInput(c client.Client, ctx context.Context, mapInput *codecTypes.AddMapConfigInput, hz *hazelcastv1alpha1.Hazelcast, m *hazelcastv1alpha1.Map) error {
```
SeriyBg, line 251: Just a small nit: it's common to have the `ctx context.Context` as the first argument of the function

api/v1alpha1/map_types.go
```diff
@@ -189,6 +194,48 @@ const (
...
+	// +optional
+	InitialMode InitialModeType `json:"initialMode,omitempty"`
+
+	// Name of your class implementing MapLoader and/or MapStore interface.
+	ClassName string `json:"className"`
+
+	// Number of seconds to delay the storing of entries.
+	// +kubebuilder:default:0
+	// +optional
+	WriteDelaySeconds int32 `json:"writeDelaySeconds"`
```
SeriyBg, line 209: Shouldn`t `+optional` parameters be a pointer? Otherwise, it will always have a `zero` value when not specified

controllers/hazelcast/hazelcast_resources.go
```diff
@@ -532,16 +536,21 @@ func clusterDataRecoveryPolicy(policyType hazelcastv1alpha1.DataRecoveryPolicyTy
...
+				return err
+			}
+			cfg.Map[mcfg.MapName()] = m
 		}
 	}
+	return nil
 }
 
-func createMapConfig(hz *hazelcastv1alpha1.Hazelcast, m *hazelcastv1alpha1.Map) config.Map {
+func createMapConfig(c client.Client, ctx context.Context, hz *hazelcastv1alpha1.Hazelcast, m *hazelcastv1alpha1.Map) (config.Map, error) {
```
SeriyBg, line 553: Same here for the `Context`

controllers/hazelcast/hazelcast_resources.go
```diff
@@ -560,9 +569,56 @@ func createMapConfig(hz *hazelcastv1alpha1.Hazelcast, m *hazelcastv1alpha1.Map)
...
+	return map[string]config.WanReplicationReference{
+		ref.Name: {
+			MergePolicyClassName: ref.MergePolicyClassName,
+			RepublishingEnabled:  ref.RepublishingEnabled,
+			Filters:              ref.Filters,
+		},
+	}
+}
+
+func getMapStoreProperties(c client.Client, ctx context.Context, sn, ns string) (map[string]string, error) {
```
SeriyBg, line 607: And here as well 😃

api/v1alpha1/map_types.go
mtyazici: Because the default value for `writeDelaySeconds` is also 0 which is the same as default Go value,  I did not add omitempty and pointer for the field.

controllers/hazelcast/hazelcast_resources.go
mtyazici: Alright :D, I will change the all those parts. Thanks for the suggestion :)

SeriyBg (approved):

hasancelik (approved):

PR #298, opened 2022-07-26
[CN-473] Add backup package with retry logic

dzeromski-hazelcast:
In addition to creating `backup` package we needed to move `client` and `config` to new packages to avoid circular dependency.

Retry logic will try to rerun http request only if it was successful http request but Hazelcast status is other than `success`.

### DIFF
--- a/controllers/hazelcast/hot_backup_controller.go
+++ b/controllers/hazelcast/hot_backup_controller.go
@@ -19,6 +19,7 @@ import (
 	"sigs.k8s.io/controller-runtime/pkg/reconcile"
 
 	hazelcastv1alpha1 "github.com/hazelcast/hazelcast-platform-operator/api/v1alpha1"
+	"github.com/hazelcast/hazelcast-platform-operator/internal/backup"
 	n "github.com/hazelcast/hazelcast-platform-operator/internal/naming"
 	"github.com/hazelcast/hazelcast-platform-operator/internal/upload"
 	"github.com/hazelcast/hazelcast-platform-operator/internal/util"
@@ -29,6 +30,8 @@ type HotBackupReconciler struct {
 	Log       logr.Logger
 	scheduled sync.Map
 	cron      *cron.Cron
+
+	backup map[types.NamespacedName]struct{}
 }
 
 func NewHotBackupReconciler(c client.Client, log logr.Logger) *HotBackupReconciler {
@@ -36,6 +39,7 @@ func NewHotBackupReconciler(c client.Client, log logr.Logger) *HotBackupReconcil
 		Client: c,
 		Log:    log,
 		cron:   cron.New(),
+		backup: make(map[types.NamespacedName]struct{}),
 	}
 }
 
@@ -77,7 +81,7 @@ func (r *HotBackupReconciler) Reconcile(ctx context.Context, req reconcile.Reque
 		return
 	}
 
-	if hb.Status.State.IsRunning() {
+	if hb.Status.State.IsRunning() || r.checkBackup(req.NamespacedName) {
 		logger.Info("HotBackup is already running.",
 			"name", hb.Name, "namespace", hb.Namespace, "state", hb.Status.State)
 		return
@@ -125,6 +129,7 @@ func (r *HotBackupReconciler) Reconcile(ctx context.Context, req reconcile.Reque
 			return result, err
 		}
 		r.removeSchedule(req.NamespacedName, logger)
+		r.lockBackup(req.NamespacedName)
 		go r.startBackup(context.Background(), req.NamespacedName, hazelcastName, logger) //nolint:errcheck
 	}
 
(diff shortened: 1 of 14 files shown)


internal/rest/hazelcast_service.go
```diff
@@ -0,0 +1,153 @@
...
+}
+
+type HazelcastStatus struct {
+	Status string `json:"status,omitempty"`
+
+	// exceptionResponse
+	Message string `json:"message,omitempty"`
+}
+
+func (s *HazelcastService) HotBackup(ctx context.Context, clusterName string) (*http.Response, error) {
```
mtyazici, line 86: Is there a reason we are not returning `HazelcastStatus`, we seem to do it for other functions with their respective responses.

controllers/hazelcast/hot_backup_controller.go
```diff
@@ -240,40 +236,29 @@ func (r *HotBackupReconciler) startBackup(ctx context.Context, backupName types.
...
-			logger := logger.WithValues("uuid", uuid)
+			logger := logger.WithValues("uuid", m.UUID)
 
 			logger.Info("Member status monitor started")
 			defer logger.Info("Member status monitor finished")
 
 			logger.Info("Wait for member backup to finish")
-			if err := waitUntilBackupSucceed(groupCtx, c, uuid); err != nil {
-				return err
+			if err := m.Wait(groupCtx); err != nil {
```
mtyazici, line 259: If the backup takes long time, this will block the reconciler function for all other HotBackup resources. I think it would be good to increase the `MaxConcurrentReconciler`. Example in the [link](https://kubebyexample.com/en/learning-paths/operator-framework/operator-sdk-go/controller-options).

internal/rest/hazelcast_service.go
```diff
@@ -0,0 +1,153 @@
...
+	}
+	return &HazelcastService{
+		client: &Client{
+			BaseURL: baseURL,
+			client:  &http.Client{},
+		},
+	}, nil
+}
+
+type HazelcastState struct {
```
mtyazici, line 27: Out of curiosity, this struct seems to be for the response for some Rest calls. Would it be better to name it `HazelcastStateResponse`?

mtyazici (commented):
LGTM, left a few comments mostly cosmetic.

controllers/hazelcast/hot_backup_controller.go
dzeromski-hazelcast: We already run backup in separate goroutine to not block reconciler:
```go
go r.startBackup(context.Background(), req.NamespacedName, hazelcastName, logger) //nolint:errcheck
```
https://github.com/hazelcast/hazelcast-platform-operator/blob/9506f86ad24569784d9cb5e2276110a625ffdbd0/controllers/hazelcast/hot_backup_controller.go#L129

internal/rest/hazelcast_service.go
dzeromski-hazelcast: Too long for my taste :wink:. We could do it in future when there are more types in `rest` package.

internal/rest/hazelcast_service.go
dzeromski-hazelcast: Status I would argue is a special case. We convert it to `HazelcastError` if the request returned failed status. Successful status does not convey any other information other than it is not failed/error. That is why I think returning error or nil is enough for this API.

controllers/hazelcast/hot_backup_controller.go
mtyazici: Ah, cool. I didn't pay attention to that part. Then this is OK.

mtyazici (approved):
LGTM :+1: . Btw, I ran the E2E tests on this branch for AWS and we got the same Hazelcast errors for external backup. I guess we still need to do more changes.

controllers/hazelcast/hot_backup_controller.go
```diff
@@ -29,13 +30,17 @@ type HotBackupReconciler struct {
 	Log       logr.Logger
 	scheduled sync.Map
 	cron      *cron.Cron
+
+	mu     sync.Mutex
```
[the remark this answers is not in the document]
dzeromski-hazelcast: To be honest we don't need that mutex because we have a single reconciler goroutine. I added it for extra safety. In addition we plan to remove this and refactor when we are done with fixing all e2e flaky tests.

What is the advantage of sync.Map? I have to admit that I don't like the sync.Map API, not sure if it makes things easier to understand :wink:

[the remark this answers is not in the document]
SeriyBg: The `sync.Map` is basically the map with the mutex itself, so it just hides the locks. IMO, having `sync.Map` is more cohesive than having two data structs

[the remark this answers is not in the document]
dzeromski-hazelcast: Removed the lock then. Hope you are happy :)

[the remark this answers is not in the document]
SeriyBg: But you didn't replace it with the `sync.Map`, won't we have a race condition?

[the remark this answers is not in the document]
dzeromski-hazelcast: No. Reconciler call are not concurrent in our configuration.

SeriyBg (approved):

PR #299, opened 2022-07-27
[CN-369] WAN Sync

SeriyBg:
PR closed by Hazelcast automation as no activity (>3 months). Please reopen with comments, if necessary. Thank you for using Hazelcast and your valuable contributions


controllers/hazelcast/t/main.go
```diff
@@ -0,0 +1,5 @@
+package main
```
SeriyBg, line 1: This one is accidentally here, I will remove it

controllers/hazelcast/wansync_controller.go
```diff
@@ -0,0 +1,191 @@
...
+			return updateWanSyncStatus(ctx, r.Client, wan, wanSyncFailedStatus(err))
+		} else {
+			return updateWanSyncStatus(ctx, r.Client, wan, wanSyncPendingStatus())
+		}
+	}
+
+	// Check publisherId is registered to the status, otherwise issue WanReplication config to Hazelcast
+	if wan.Status.PublisherId == "" {
+		logger.Info("Applying WAN configuration")
+		publisherId, err := r.getWanPublisherId(ctx, cli, wan)
```
mtyazici, line 82: My question is does this not start a Wan Replication right away if they give the config in WanSync? And then we send a wan sync request later, what happens in between. Up until this point it seems like what we are doing is the same as Wan Replication.

mtyazici (commented):
Thanks for the PR, I have some questions regarding to the WanSync and reconciler logic order. Also left a few tirivial comments.

controllers/hazelcast/hazelcast_rest_client.go
```diff
@@ -137,6 +139,47 @@ func (c *RestClient) HotBackup(ctx context.Context) error {
...
+	if s := rBody["status"]; s != "success" {
+		return fmt.Errorf("error executing WAN Sync: %s, %s", s, rBody["message"])
+	}
+	return nil
+}
+
+func (c *RestClient) wanSyncData(wan WanPublisherObject) string {
+	wpc := wan.WanPublisherConfig()
+	if wpc.MapResourceName != "" {
+		return fmt.Sprintf("%s&&%s&%s&%s",
```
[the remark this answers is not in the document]
SeriyBg: Yeap, this is for password which is not supported for now in operator

controllers/hazelcast/wansync_controller.go
SeriyBg: That's true. For WanSync we need to use a WAN publisher. The only difference is that the Sync operation will send the existing events to the target cluster, while the replication is only working with the new events

PR #300, opened 2022-07-27
[CN-444] Add Executor Service Support

cagric0:
Add executor service config to hz spec and configmap
Add codecs for executor service dynamic config


internal/config/hazelcast.go
```diff
@@ -103,6 +106,24 @@ type WanReplicationReference struct {
...
 
+type ExecutorService struct {
+	PoolSize      int32 `yaml:"pool-size"`
+	QueueCapacity int32 `yaml:"queue-capacity"`
+}
+
+type DurableExecutorService struct {
+	PoolSize   int32 `yaml:"pool-size"`
+	Durability int32 `yaml:"durability"`
+	Capacity   int32 `yaml:"capacity"`
```
SeriyBg, line 117: How abut extracting and resuing this as a separate `inline` object? It will avoid duplications

api/v1alpha1/hazelcast_types.go
```diff
@@ -86,8 +86,85 @@ type HazelcastSpec struct {
 	// Custom Classes to Download into Class Path
 	// +optional
 	CustomClass *CustomClassConfiguration `json:"customClass,omitempty"`
+
+	// +optional
+	ExecutorServices *ExecutorServices `json:"executorServices,omitempty"`
```
SeriyBg, line 91: Just thinking about the API, for now it will be like this:
```yaml
executorServices:
  basicExecutorServices:
  - name: serviceBasic-1
    poolSize: 3
  - name: serviceBasic-2
    poolSize: 3
  durableExecutorServices:
  - name: serviceDurable-1
    poolSize: 3
    durability: 2
  - name: serviceDurable-2
    poolSize: 3
    durability: 2
  scheduledExecutorServices: []
```
What if, instead we will eliminate one level, and make it similar to the `Volumes` in `PodSpec`?
```yaml
executorServices:
  - name: serviceBasic-1
    basicExecutorService:
      poolSize: 3
  - name: serviceDurable-2
    durableExecutorService:
      durability: 2
      poolSize: 3
  - name: serviceScheduled-3
    scheduledExecutorService:
      capacityPolicy: PER_NODE
```

hasancelik: After a small discussion with @cagric0, I understand that all types of executor-services are independent of each other. So we can keep their config separated as well:

```
executorServices:
  - name: service-1
    poolSize: 3
durableExecutorServices:
  - name: serviceDurable-1
    poolSize: 3
    durability: 2
scheduledExecutorServices: []
```

Also, there is no `basic executor-service` term at the Hazelcast side. It might be confusing for users.

internal/config/hazelcast.go
cagric0: My first intention was using inline object but then I realize when we use inline object, we need to use inline object's name explicitly which is confusing the code reader. That's why I decided not to use inline

hasancelik (approved):
LGTM 👍

controllers/hazelcast/hazelcast_resources.go
```diff
@@ -1089,6 +1129,7 @@ func env(h *hazelcastv1alpha1.Hazelcast) []v1.EnvVar {
 			Value: strconv.FormatBool(util.IsPhoneHomeEnabled()),
 		},
 		{
+
```
SeriyBg, line 1132:
```suggestion
```

SeriyBg (approved):
LGTM

PR #301, opened 2022-07-29
[CN-474] Change hotBackup retry count from 4 to 8

dzeromski-hazelcast:
Changed steps count and factor. In total it should take the same amount of time.

### DIFF
--- a/internal/backup/backup.go
+++ b/internal/backup/backup.go
@@ -17,6 +17,14 @@ import (
 	hzconfig "github.com/hazelcast/hazelcast-platform-operator/controllers/hazelcast/config"
 )
 
+// backupBackoff is based on retry.DefaultBackoff
+var backupBackoff = wait.Backoff{
+	Steps:    8,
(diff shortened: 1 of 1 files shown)


mtyazici (commented):
Looking at the descriptions of the fields, does changing `factor` to 2 times not change the duration to 2 times more at each step. So, this configuration should not take the same amount, am I missing something?

dzeromski-hazelcast:
> Looking at the descriptions of the fields, does changing `factor` to 2 times not change the duration to 2 times more at each step. So, this configuration should not take the same amount, am I missing something?

Worst case scenario 640ms vs previous 250ms. Close enough for me :P

mtyazici:
That is OK by me, I thought you were increasing the retries but decreasing the waiting time in-between because you said `it should take the same amount of time.`. I am fine with increasing the retry times and interval.

mtyazici (approved):

SeriyBg (approved):

PR #304, opened 2022-08-01
[CN-445] Add support for custom class from configmaps

mtyazici:
Changes:
- Added ConfigMap support to Custom Class feature.
- Refactored StatefulSet reconciliation function to centralize creation of `initContainer`, `volumes` and `volumeMounts` fields. This allows us to directly use these functions in the `createOrUpdate` function without any if-else check in the `createOrUpdate` function.
- Reordered functions


controllers/hazelcast/hazelcast_resources.go
```diff
@@ -783,6 +783,17 @@ func backupAgentContainer(h *hazelcastv1alpha1.Hazelcast) v1.Container {
 	}
 }
 
+func initContainers(h *hazelcastv1alpha1.Hazelcast) []corev1.Container {
+	var ctnrs []corev1.Container
```
dzeromski-hazelcast, line 787: Why `ctnrs` and not just  `containers`? :wink:

controllers/hazelcast/hazelcast_resources.go
```diff
@@ -1026,6 +1061,19 @@ func env(h *hazelcastv1alpha1.Hazelcast) []v1.EnvVar {
 	return envs
 }
 
+func javaClassPath(h *hazelcastv1alpha1.Hazelcast) string {
+	b := []string{n.CustomClassBucketPath + "/*"}
+	if !h.Spec.CustomClass.IsConfigMapEnabled() {
```
dzeromski-hazelcast, line 1066: I think this is not needed: `strings.Join([]string{"a"}, ":")` returns `a` (https://go.dev/play/p/YWtbx-aUu4Y).

dzeromski-hazelcast:
Great work with refactoring!

api/v1alpha1/hazelcast_types.go
```diff
@@ -280,8 +287,13 @@ func (c *ExposeExternallyConfiguration) IsEnabled() bool {
 }
 
 // Returns true if customClass configuration is specified.
-func (c *CustomClassConfiguration) IsEnabled() bool {
-	return c != nil && !(*c == (CustomClassConfiguration{}))
+func (c *CustomClassConfiguration) IsBucketEnabled() bool {
+	return c != nil && c.BucketConfiguration != nil
+}
+
+// Returns true if customClass configuration is specified.
```
cagric0, line 294: A clearer comment would be better.

cagric0 (commented):
LGTM except one issue. What if BucketConfiguration and Configmaps are given under CustomClassConfiguration at the same time? I think we are not checking it and It can cause some problems.

controllers/hazelcast/hazelcast_resources.go
mtyazici: I am not sure what is not needed, after the if check we are adding into the slice for configmaps. Can you share a snippet of how the function should change?

mtyazici:
@cagric0, users can give both configmaps and bucket configurations  at the same time, they are not mutually exclusive. So we do not need to check for it.

api/v1alpha1/hazelcast_types.go
mtyazici: Sure, I will update it.

cagric0 (approved):

controllers/hazelcast/hazelcast_resources.go
dzeromski-hazelcast: Ah now I see you negate first condition. Well then maybe just remove: `if !h.Spec.CustomClass.IsConfigMapEnabled() {` condition?

dzeromski-hazelcast (approved):

dzeromski-hazelcast:
Just to clarify do we need to worry about CustomClass.TriggerSequence in case of configmaps?

mtyazici:
Files are updated by kubelet but Hazelcast won't pick up the changes because the classes are loaded at the startup AFAIK.

controllers/hazelcast/hazelcast_resources.go
mtyazici: When we remove the condition it results in nil pointer dereference

mtyazici:
Flaky test not related to the PR failed, I will merge the PR.

PR #315, opened 2022-08-09
[CN-491] Add new relic logging for backup-e2e-gke workflow

dzeromski-hazelcast:


.github/workflows/backup-e2e-gke.yaml
```diff
@@ -35,6 +35,44 @@ jobs:
             --num-nodes=2
           sleep 30
 
+  new-relic-setup:
+    name: Setup New Relic agent
+    needs: [create-gke-cluster]
+    env:
+      NR_CLUSTER_NAME: gke-operator
```
SeriyBg, line 42: It seems that this env var isn't used anywhere. The idea was that it should be used here `--set "global.cluster=gke-operator"` but somehow I've missed it 😃 
Besides that, does it make sense to set a different name to distinguish this workflow from the common nightly tests, wdyt?

SeriyBg (approved):

cheels (approved):
LGTM 👍

PR #318, opened 2022-08-09
[CN-484] stopping hotbackup gracefully

semihbkgr:


controllers/hazelcast/hot_backup_controller.go
```diff
@@ -174,6 +177,8 @@ func (r *HotBackupReconciler) executeFinalizer(ctx context.Context, hb *hazelcas
 		Name:      hb.Name,
 		Namespace: hb.Namespace,
 	}
+	r.cancelMap[key]()
```
mtyazici, line 180: This can panic if the cancel function is not put into the map yet. I think it is better to check if the key exists first and then call the function. It can happen if somehow the backup failed before reaching that part in the code.

controllers/hazelcast/hot_backup_controller.go
```diff
@@ -174,6 +177,10 @@ func (r *HotBackupReconciler) executeFinalizer(ctx context.Context, hb *hazelcas
 		Name:      hb.Name,
 		Namespace: hb.Namespace,
 	}
+	if cancelFunc, ok := r.cancelMap[key]; ok {
+		cancelFunc()
```
SeriyBg, line 181: I'm a bit confused, as I don't see what exactly this cancelation does. It will cancel the context, but will it interrupt any work in progress?

controllers/hazelcast/hot_backup_controller.go
```diff
@@ -30,16 +30,17 @@ type HotBackupReconciler struct {
...
 func NewHotBackupReconciler(c client.Client, log logr.Logger) *HotBackupReconciler {
 	return &HotBackupReconciler{
-		Client: c,
-		Log:    log,
-		cron:   cron.New(),
-		backup: make(map[types.NamespacedName]struct{}),
+		Client:    c,
+		Log:       log,
+		cron:      cron.New(),
+		cancelMap: make(map[types.NamespacedName]context.CancelFunc),
```
dzeromski-hazelcast, line 42: Can we have single map with struct for cancelFunc and everything else instead of two maps?

controllers/hazelcast/hot_backup_controller.go
SemihBKGR: Yes, it interrupts. When the context which is passed to startBackup function is cancelled, it notifies the members(agents) to cancel backing up process.

``` go
func (r *HotBackupReconciler) startBackup(ctx context.Context, backupName types.NamespacedName, hazelcastName types.NamespacedName, logger logr.Logger) (ctrl.Result, error) {

...

    if err := u.Wait(groupCtx); err != nil {
        if errors.Is(err, context.Canceled) {
	    // notify agent so we can cleanup if needed
	    logger.Info("Cancel upload")
	    return u.Cancel(ctx)
        }
        return err
    }
```

SeriyBg: Ok, I see. Can we add this check to the unit test? We can check the `cancel` function on the `backup.NewClusterBackup` is called

dzeromski-hazelcast (approved):

controllers/hazelcast/hot_backup_controller.go
SemihBKGR: We can do it. But in this time, we would have to change some checks (for scheduled backup) and it becomes more complex. Also the purpose of these two maps are different from each other.

controllers/hazelcast/hot_backup_controller.go
SemihBKGR: I am writing e2e test for testing `backup.NewClusterBackup`, because in unit it is required to mock hazelcast client.

controllers/hazelcast/hot_backup_controller.go
SeriyBg: @SemihBKGR I think we can easily use the `context.CancelFunc` instead of the empty `struct` for all the checks. The purpose is just to check that the backup is present, and having the `CancelFunc` can serve this purpose

semihbkgr: `backup` map is just used for instant backup (not scheduled backup. However `cancelMap` is used for both instant and scheduled backups. As I said, we can use just one common map, but in this time we have to change some checks.

```go
cancelCtx, cancelFunc := context.WithCancel(context.Background())
// cancelMap is used for both instant and scheduled backups
r.cancelMap[req.NamespacedName] = cancelFunc
if hb.Spec.Schedule != "" {
  logger.Info("Adding backup to schedule")
  r.scheduleBackup(cancelCtx, hb.Spec.Schedule, req.NamespacedName, hazelcastName, logger)
} else {
  result, err = r.updateStatus(ctx, req.NamespacedName, hbWithStatus(hazelcastv1alpha1.HotBackupPending))
  if err != nil {
	  return result, err
  }
  r.removeSchedule(req.NamespacedName, logger)
  // backup map is only used for instant backup
  r.lockBackup(req.NamespacedName)
  go r.startBackup(cancelCtx, req.NamespacedName, hazelcastName, logger) 
}

```

test/e2e/hazelcast_backup_slow_test.go
```diff
@@ -304,4 +304,77 @@ var _ = Describe("Hazelcast Backup", Label("backup_slow"), func() {
...
+			DiscoveryServiceType: corev1.ServiceTypeLoadBalancer,
+			MemberAccess:         hazelcastcomv1alpha1.MemberAccessLoadBalancer,
+		}
+		hazelcast.Spec.Resources = &corev1.ResourceRequirements{
+			Limits: map[corev1.ResourceName]resource.Quantity{
+				corev1.ResourceMemory: resource.MustParse(strconv.Itoa(mapSizeInGb) + "Gi")},
+		}
+		hazelcast.Spec.Persistence.Pvc.RequestStorage = &[]resource.Quantity{resource.MustParse(strconv.Itoa(mapSizeInGb) + "Gi")}[0]
+		CreateHazelcastCR(hazelcast)
+		defer func() {
```
mtyazici, line 332: Cleanup happens after the test is finished we don't need this defer function

test/e2e/hazelcast_backup_slow_test.go
```diff
@@ -304,4 +304,77 @@ var _ = Describe("Hazelcast Backup", Label("backup_slow"), func() {
...
+
+		By("delete hotbackup to cancel backup process")
+		err = k8sClient.Delete(ctx, hb)
+		Expect(err).ToNot(HaveOccurred())
+
+		// hazelcast logs
+		hzLogs := InitLogs(t, hzLookupKey)
+		defer hzLogs.Close()
+		scanner := bufio.NewScanner(hzLogs)
+		test.EventuallyInLogs(scanner, 10*Second, logInterval).Should(ContainSubstring("Starting new hot backup with sequence"))
```
mtyazici, line 370: Maybe, we can put an assertion to see Hazelcast backup process is done.

test/e2e/helpers_test.go
```diff
@@ -73,6 +73,18 @@ func InitLogs(t Time, lk types.NamespacedName) io.ReadCloser {
 	return logs
 }
 
+func InitAgentLogs(t Time, lk types.NamespacedName) io.ReadCloser {
```
mtyazici, line 76: I think it is better to name it `sidecarAgentLogs` since `backup-agent` is not an init container.

controllers/hazelcast/hot_backup_controller.go
SeriyBg: Oh, I see. But it seems that we can use the `backup` map for both backups as well, it won't break anything and we will get rid of one `map`

mtyazici (approved):

PR #334, opened 2022-08-18
[CN-411] Expose inMemoryFormat in Map CR and properties in HZ CR

cagric0:


api/v1alpha1/hazelcast_types.go
```diff
@@ -95,6 +95,9 @@ type HazelcastSpec struct {
 
 	// +optional
 	ScheduledExecutorServices []ScheduledExecutorServiceConfiguration `json:"scheduledExecutorServices,omitempty"`
+
+	// +optional
+	Properties map[string]string `json:"properties,omitempty"`
```
SeriyBg, line 100: As far as I understand, with the `Properties` we are exposing the raw Hazelcast configs to the user. Do we really want to do so?

cagric0: Yes, you are right about your concern but we already talked about it with Hasan. We need to expose some of the properties. As you can see in validation.go, we have blacklist of properties and we are filtering the properties according to that blacklist. For now the blacklist is empty, but later we are gonna add some properties to that list to prevent them to be exposed.

test/integration/hazelcast_controller_test.go
```diff
@@ -471,6 +473,38 @@ var _ = Describe("Hazelcast controller", func() {
 		})
 	})
 
+	Context("Hazelcast CustomResource with properties", func() {
+		FIt("should pass the values to ConfigMap", Label("fast"), func() {
```
mtyazici, line 477: Forgotten focused test

mtyazici (changes requested):
Thanks for the PR! Could you also update the `createMapConfig` function in `hazelcast_resources.go` to reflect the changes done to `InMemoryFormat`.

mtyazici (approved):
LGTM :+1:

SeriyBg (approved):

PR #335, opened 2022-08-18
Added direct link to NewRelic using CR_ID

cheels:


cagric0 (approved):
Thanks for the improvement on the report! 
I manually tested and it worked. We can directly see CR_ID related logs in NewRelic.

PR #336, opened 2022-08-19
[CN-517] Implement Phone Home as separate goroutine

mtyazici:
With the changes, `Hazelcast` and `ManagementCenter` do not collect any metrics in the reconciler function. They only trigger the Phone Home collection function.

To be able to collect metrics only in a separate goroutine I removed the `AverageClusterCreationLatency`, `AverageMCCreationLatency` metrics from the `PhoneHomeData`.


controllers/hazelcast/hazelcast_controller.go
```diff
@@ -36,16 +35,16 @@ type HazelcastReconciler struct {
 	Log                  logr.Logger
 	Scheme               *runtime.Scheme
 	triggerReconcileChan chan event.GenericEvent
-	metrics              *phonehome.Metrics
+	phTrigger            chan struct{}
```
SeriyBg, line 38: Just curious why you decided to use a shared `chan` instead of having a separate reconciler? I think with a separate reconciler we can decouple the `Hazelcast/MC` ones from Phone Home at all and avoid using `chan`. Performance should not be the case, IMO

mtyazici: Performance should not  be a problem as you said but that reconciler will be triggered after any change to the any of the Custom Resources we have but we only care about the time when they are created successfully. I beliebe it is also a bit out of the scope for a reconciler. Don't we need to create a dummy custom resource for that one?

SeriyBg: It will get triggered. However, the same happens with this reconciler as well, doesn't it?
I don't think we need a dummy CR, as the separate reconciler can also listen to the needed CR (Hazelcast and MC in this case)

mtyazici: > However, the same happens with this reconciler as well, doesn't it?
Not sure if I got that one, which reconciler are you talking about?

Also how are we going to handle the trigger mechanism if we choose to use a separate reconciler? We will not know which resource got the reconciler triggered. Are we going to send Phone Home directly when the reconciler gets triggered? Then, we would be sending Phone Home for any event.

SeriyBg: We don't need to send the PhoneHome call directly, we can just collect the metrics on the reconcile call, as it is done now, and on a scheduled basis, send the call to the PH server.
And it seems that we can use `Predicates` to distinguish not only the object type but also the event type (distinguishing create from the update, etc.)
And it seems we can also use this for other reconcilers like `HotBackup` etc.

SeriyBg (approved):

controllers/hazelcast/hazelcast_controller.go
```diff
@@ -205,6 +190,10 @@ func (r *HazelcastReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 		r.addExecutorServices(ctx, hzClient.Client, newExecutorServices)
 	}
 
+	if util.IsPhoneHomeEnabled() && !util.IsSuccessfullyApplied(h) {
+		go func() { r.phTrigger <- struct{}{} }()
```
dzeromski-hazelcast, line 194: Maybe use buffered channel?

dzeromski-hazelcast (approved):

controllers/hazelcast/hazelcast_controller.go
mtyazici: I am actually using a buffered channel, I added this just in case the buffer becomes full. Maybe I can increase the buffer size and remove the goroutine calls.

PR #340, opened 2022-08-22
[CN-454] Add User Code Deployment Hazelcast config

mtyazici:


api/v1alpha1/hazelcast_types.go
```diff
@@ -194,11 +194,14 @@ type BucketConfiguration struct {
 
 // UserCodeDeploymentConfig contains the configuration for User Code download operation
 type UserCodeDeploymentConfig struct {
+	// When true, allows user code deployment from clients, ConfigMaps and buckets.
+	Enabled bool `json:"enabled"`
```
SeriyBg, line 198: Do we need an explicit Enabled field? Usually, we can imply it from the configuration itself. IMO, it would be good to be consistent in the API

mtyazici: Normally, as you said we do not do that. However, there is a `user-code-deployment` config in Hazelcast side which allows clients to deploy classes dynamically. The problem here is that, in `userCodeDeployment` field we allow deploying classes from buckets and configmaps. But what happens if the user just wants to enable `user-code-deployment` in the Hazelcast side without deploying classes with bucket and configmap? Although, I am not really OK with the solution this is the only one seemed reasonable enough. If you have a suggestion, please comment :)

SeriyBg: I see. Well, one solution would be for the user to declare `userCodeDeployment: {}`, but it's not very intuitive. Could we just always enable it by default? Is there a reason why it should be disabled, maybe some security or something?

mtyazici: I am not sure, maybe instead of enabled field, we can give a field like `clientCodeDeploymentEnabled`. It is hard to shorten the name though :D

SeriyBg: Well, another option could be allowing users to explicitly disable it by having a `disabled` field. So by default, it will be enabled, but if one wants to disable it, one needs to configure the following:
```yaml
userCodeDeployment:
  disabled: true
```
However, I'm not sure this is a better option than just having the `Enabled` field.

While writing this comment, I've come up with another idea. What about making the `Enabled` field `true` by default? By that, I mean that if users don't provide the `userCodeDeployment` config at all, it will be considered disabled. If users want just to enable it, one could set the `enable: true`. However, if users want to configure `BucketConfiguration`, he does not need to set `enable: true`, just the `BucketConfiguration`. WDYT?

cagric0: In my opinion, the best option is adding another field (`clientCodeDeploymentEnabled`) like Muhammet suggested. We have three options for userCodeDeployment and except client, the other two options have their own configuration. Additional enabled field for those two options looks confusing. I think, we should have separate enabled field for client.

mtyazici: This is a good solution but I think then the `enabled` field does not really execute its meaning. I think a different name might be a better solution if we choose to implement that logic.

SeriyBg: @cagric0 but this way, we will have `userCodeDeployment` and `clientCodeDeploymentEnabled` on the same level of the configuration in `Hazelcast` CR. IMO this will lead to confusion

cagric0: No, actually I was thinking to put `clientCodeDeploymentEnabled` field under `userCodeDeployment`.

SeriyBg: Yes, but this can cause another confusion: can one configure the `userCodeDeployment` from the CR but disable it for the client via `clientCodeDeploymentEnabled `? 😃

cagric0 (approved):
LGTM!

SeriyBg (approved):
LGTM 👍

PR #341, opened 2022-08-23
[CN-514] Add Phone Home for Map CR

mtyazici:


controllers/hazelcast/map_controller.go
```diff
@@ -33,8 +33,18 @@ import (
 // MapReconciler reconciles a Map object
 type MapReconciler struct {
 	client.Client
-	Log    logr.Logger
-	Scheme *runtime.Scheme
+	Log              logr.Logger
+	Scheme           *runtime.Scheme
+	phoneHomeTrigger chan struct{}
```
SeriyBg, line 38: IMO this channel is redundant, as we can calculate the map counts before sending the data to the PH server

SeriyBg (changes requested):

internal/phonehome/phonehome.go
```diff
@@ -190,3 +198,28 @@ func (phm *PhoneHomeData) fillMCMetrics(cl client.Client) {
...
+	persistedMapCount := 0
+	mapStoreMapCount := 0
+
+	ml := &hazelcastv1alpha1.MapList{}
+	err := cl.List(context.Background(), ml, client.InNamespace(os.Getenv(n.NamespaceEnv)))
+	if err != nil {
+		return //TODO maybe add retry
+	}
+
+	for _, m := range ml.Items {
```
SeriyBg, line 213: I believe the tests are failing for the same reason as I have encountered. You need to add a check for `ml.Items` no being `nil`.

controllers/hazelcast/map_controller.go
mtyazici: But this can cause the created map to not be sent before the next ticker comes which happens every 24 hours. I thought our intention was to send a phone home after each CR is created.

SeriyBg: Doesn't it make the ticker schedule redundant, then?

mtyazici: We do not send phone homes for the updates done to the CRs. Those changes are sent by the timer.

SeriyBg: But we do send it on every other CR creation, don't we? I mean, when HotBackup CR is created, it will also trigger the PhoneHome

cagric0: I agree with Muhammet since if we do not use channel then we can lose some of the creation. For example, before the next 24th hour, user can delete all created maps.

And also about the ticker, again I think we should keep it. The ticker won't be helpful for wan or map since the ph data for them are not updatable but for HZ and MC there could be some ph fields which are updatable. Either we should send on update of each CR or use ticker. Since sending ph data on every update would be too much, I think we should keep ticker. 

Last but not least, I think we should be consistent on each CR about using phoneHomeTrigger. It would be easy to understand the ph mechanism and also in any case we shouldn't lose any creation even if it's not highly possible. I mean your PR @SeriyBg , I saw that you didn't use channel, I propose to use it anyway. Probably we won't lose any creation of WAN since user won't delete it frequently but still nice to cover this case.

mtyazici: Currently there is no trigger for that one because we did not add hotbackup information to phone home, but that is the idea. There is also a time windows of 24 hours, where users can create and delete some CRs and we would not know about them because we do not have triggers for CR creations.

SeriyBg (approved):
You convinced me 😄

cagric0 (approved):
LGTM! but I have a question.
Shouldn't we update the test or add new one for map?

cagric0:
> > Shouldn't we update the test or add new one for map?
> 
> I think we should, but the problem with adding more tests is there will be more rows in the production table. I think we should only have one test where we send only one row. Not sure if it should be done in this PR though. But I can add tests for Map CR regardless.

I see your point and you are right. I think we should discuss it with the team. Then maybe we can create a task for updating PH test.

PR #362, opened 2022-09-09
[CN-532] Add CronHotbackup CR

mtyazici:
Changes:
- Remove schedule from HotBackup
- Add CronHotbackup resource
- Add integration and e2e tests


controllers/hazelcast/cronhotbackup_controller.go
```diff
@@ -0,0 +1,357 @@
...
+	}
+	return nil
+}
+
+func (r *CronHotBackupReconciler) executeFinalizer(ctx context.Context, chb *hazelcastv1alpha1.CronHotBackup) error {
+	if !controllerutil.ContainsFinalizer(chb, n.Finalizer) {
+		return nil
+	}
+	r.removeSchedule(types.NamespacedName{Name: chb.Name, Namespace: chb.Namespace})
+	err := r.deleteDependentHotBackups(ctx, chb)
```
dzeromski-hazelcast, line 142: I wonder if it is a good idea to delete hotbackups when cronbackup is deleted. Users may accidentally remove all backups if for example they just wanted to adjust cronbackup by removing and adding CR.

Its not a blocker for review. Just wanted us to think about it.

dzeromski-hazelcast (approved):

SeriyBg (approved):

controllers/hazelcast/cronhotbackup_controller.go
mtyazici: I added the `suspend` field as we discussed offline.

PR #365, opened 2022-09-15
Cleaned up and improved the existing workflows

cheels:
* The secrets in workflow_call becomes inheritable from the master job
* Added create cluster dependency for report creation step


dzeromski-hazelcast (approved):

hasancelik (commented):
👍

PR #371, opened 2022-09-29
[CN-531] Add Topic CR

mtyazici:


cagric0:
I was doing refactor on finalizer logic in the hazelcast_resource but I encounter some issues and for now I skip the refactoring on finalizer. You can directly add dependent deletion for Topic as we did for other CRs.

cagric0 (approved):
LGTM!

hasancelik (approved):

PR #375, opened 2022-09-30
Added option to load the docker image into Kind cluster

cheels:


dzeromski-hazelcast (approved):

config/manager/update_image_pull_policy.yaml
```diff
@@ -0,0 +1,3 @@
+- op: replace
+  path: "/spec/template/spec/containers/0/imagePullPolicy"
+  value: IfNotPresent
```
dzeromski-hazelcast, line 3: I assume you added this because we have always pull option set. Maybe we should change it in operator?

cheels: @dzeromski-hazelcast Yes, exactly. And since I'm loading the image locally into the cluster with a fake tag, then with the "always" policy, it will always fail because it doesn't exist in the registry. 
As for the operator, I don't think we should change this default behavior since it's better to download a fresh image from the registry.

PR #403, opened 2022-10-17
[CN-462] Add ClearWanQueue function into WanReplication finalizer executor

mtyazici:


dzeromski-hazelcast (approved):

hasancelik (approved):
@mtyazici, is it possible to add an E2E test for this functionality?

mtyazici:
@hasancelik

Unfortunately, it is not easy to test this case without a complicated setup and event then I believe the tests would be quite flaky.

PR #408, opened 2022-10-19
[CN-282] Add entry listener support for Map CR

SeriyBg:


test/util.go
```diff
@@ -64,18 +64,19 @@ func SpecLabelsChecker() {
 		buf := make([]byte, 0, 2097152)
 		scanner.Buffer(buf, 5242880)
 
-		lbl, lblErr := regexp.Compile(`(^(.*?)It|^(.*?)Entry)(.*?)Label\((.*?)$`)
-		noLbl, noLblErr := regexp.Compile(`((^(.*?)It)(.*?)func(.*?){$)|((^(.*?)Entry)(.*?),$)`)
+		lbl, lblErr := regexp.Compile(`(^(.*?)\s+It|^(.*?)\s+Entry\()(.*?)Label\((.*?)$`)
+		noLbl, noLblErr := regexp.Compile(`((^(.*?)\s+It)(.*?)func(.*?){$)|((^(.*?)\s+Entry\()(.*?),$)`)
```
SeriyBg, line 68: The reviewer might woned what is the purpose of these changes. Here is the answer: these regexes matched the class name `org.example.SampleEntryListener` and detected it as a test declaration. To avoid it, I've changed the regex so that instead of matching `Entry`, it will match `Entry(`, also I have added the whitespace for the `It` and `Entry` matching, as in the test declaration it is always followed by indentation.

api/v1alpha1/map_types.go
```diff
@@ -63,6 +63,40 @@ type MapSpec struct {
...
+}
+
+type EntryListenerConfiguration struct {
+	// ClassName is the fully qualified name of the class that implements any of the Listener interface.
+	// +kubebuilder:validation:MinLength:=1
+	ClassName string `json:"className"`
+
+	// IncludeValues is an optional attribute that indicates whether the event will contain the map value.
+	// +kubebuilder:default:=true
+	IncludeValues *bool `json:"includeValues"`
```
mtyazici, line 81: If `IncludeValues` is not optional, we do not need a default value in kubebuilder command. Should we make it optional?

api/v1alpha1/map_types.go
```diff
@@ -63,6 +63,40 @@ type MapSpec struct {
...
+
+	// Local is an optional attribute that indicates whether the map on the local member can be listened to.
+	// +kubebuilder:default:=false
+	Local *bool `json:"local"`
```
mtyazici, line 85: If `local` is not optional, we do not need a default value in kubebuilder command. Should we make it optional?

mtyazici (commented):
LGTM in general, just left a few comments

api/v1alpha1/map_types.go
mtyazici: To keep it consistent with how we approach other fields, I would suggest we do something like this

```suggestion
        // +kubebuilder:default:=true
	IncludeValues bool `json:"includeValues,omitempty"`
```

I removed the pointer because the field has a default value and added omitempty because default go value and its real default value are not aligned. So, if the resource is created from the operator using this struct, it will have the correct default value.

At the end I would suggest we delete `GetIncludedValue()` function, and just use the field to get its value. WDYT?

api/v1alpha1/map_types.go
mtyazici: To keep it consistent with how we approach other fields, I would suggest we do something like this

```suggestion
        // +kubebuilder:default:=false
	Local bool `json:"local"`
```

I removed the pointer because the field has a default value and did not add omitempty tag because default go value and it real value are aligned. 

At the end I would suggest we delete `GetLocal()` function, and just use the field to get its value. WDYT?

mtyazici (approved):
Works both way but I added my comments for the mentioned fields again. Not a blocker though.

cagric0 (approved):
LGTM!

api/v1alpha1/map_types.go
mtyazici: We are not forcing users to specify the value because we added `+optional` comment along with default kubebuilder tags the fields will be filled in the apiserver side. 

But you are right on the omitempty part. I forgot about user giving `includeValues` as false. In this case the pointer is required for this field. But I still believe the other field does not need to be a pointer.

PR #409, opened 2022-10-20
Remove agent port from the discovery service

mtyazici:
When agent is added to the Hazelcast member pods, Kubernetes discovery plugin adds `member-ip:5701` and `member-ip:8080` into member endpoints instead of only `member-ip:5701`. This in turn becomes a problem when the plugin tries to enrich members with external IPs. It can no longer map endpoints to external IPs one to one, so no member gets an external IP. 

With the changes, this problem is fixed.


dzeromski-hazelcast (approved):

hasancelik (approved):
LGTM 👍

PR #421, opened 2022-10-28
Update deprecated set-output commands

cheels:
Starting 1st June 2023, workflows using set-output commands via stdout will fail with an error. Currently, we have a lot of warnings in our workflows


SeriyBg (approved):
Thank you for this update! Finally, the GH Actions output syntax is simplified!😄

PR #425, opened 2022-11-03
[CN-563] Refactor Hazelcast client and service structure

mtyazici:
Now all CR controllers interact with the client using a client manager and all services interact with Hazelcast client through an interface


internal/hazelcast-client/client.go
```diff
@@ -2,42 +2,63 @@ package client
...
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-	"k8s.io/apimachinery/pkg/types"
-	"sigs.k8s.io/controller-runtime/pkg/event"
-
-	hazelcastv1alpha1 "github.com/hazelcast/hazelcast-platform-operator/api/v1alpha1"
-	"github.com/hazelcast/hazelcast-platform-operator/internal/protocol/codec"
-	codecTypes "github.com/hazelcast/hazelcast-platform-operator/internal/protocol/types"
 )
 
+type ClientI interface {
```
dzeromski-hazelcast, line 13: Can we rename it to `Client` and name interface implementation something more specific like "HazelcastClient`?

internal/hazelcast-client/client_manager.go
```diff
@@ -0,0 +1,45 @@
...
+	errNoClient = errors.New("Client is not created yet")
+)
+
+type ClientManager struct {
+	Clients sync.Map
+}
+
+func (cs *ClientManager) CreateClient(ctx context.Context, h *hazelcastv1alpha1.Hazelcast) (ClientI, error) {
+	ns := types.NamespacedName{Namespace: h.Namespace, Name: h.Name}
+	client, err := cs.GetClient(ns)
```
dzeromski-hazelcast, line 22: Can we make CreateClient "synchronous"? I think it would simplify things if GetClient started the client and failed if the client didnt start. This would simplify error handling I think.

internal/hazelcast-client/client.go
```diff
@@ -2,42 +2,63 @@ package client
...
+	Running() bool
```
dzeromski-hazelcast, line 14: Can we remove it from interface and fail other functions if client is not running?

internal/hazelcast-client/client.go
```diff
@@ -2,42 +2,63 @@ package client
...
+	IsClientConnected() bool
```
dzeromski-hazelcast, line 15: Same here. Can we remove it from interface and fail other functions if client is not connected?

internal/hazelcast-client/client.go
```diff
@@ -2,42 +2,63 @@ package client
...
+	AreAllMembersAccessible() bool
```
dzeromski-hazelcast, line 16: Can we remove it from interface and fail other functions if members are not accessible?

internal/hazelcast-client/client.go
```diff
@@ -2,42 +2,63 @@ package client
...
+
+	OrderedMembers() []cluster.MemberInfo
```
dzeromski-hazelcast, line 18: Can we use "tell dont ask" rule and instead of returning Members just replace this with method `InvokeOnMembers`?

https://martinfowler.com/bliki/TellDontAsk.html

internal/hazelcast-client/client.go
```diff
@@ -2,42 +2,63 @@ package client
...
+	InvokeOnMember(ctx context.Context, req *proto.ClientMessage, uuid hztypes.UUID, opts *proto.InvokeOptions) (*proto.ClientMessage, error)
+	InvokeOnRandomTarget(ctx context.Context, req *proto.ClientMessage, opts *proto.InvokeOptions) (*proto.ClientMessage, error)
```
dzeromski-hazelcast, line 20: Do we need separate methods `InvokeOnMember` and `InvokeOnRandomTarget`? Can we replace it with one function?

internal/hazelcast-client/client.go
```diff
@@ -2,42 +2,63 @@ package client
...
+
+	Shutdown(ctx context.Context) error
```
dzeromski-hazelcast, line 22: Can we remove it from interface assuming client will be "synchronous`?

internal/hazelcast-client/client_manager.go
```diff
@@ -0,0 +1,45 @@
...
+
+var (
+	errNoClient = errors.New("Client is not created yet")
+)
+
+type ClientManager struct {
+	Clients sync.Map
+}
+
+func (cs *ClientManager) CreateClient(ctx context.Context, h *hazelcastv1alpha1.Hazelcast) (ClientI, error) {
```
dzeromski-hazelcast, line 20: Do we need a separate public method that creates client? Can we hide it and move it to GetClient and call it if the client is not found?

internal/hazelcast-client/client_manager.go
```diff
@@ -0,0 +1,45 @@
...
+	}
+	c, err := NewClient(ctx, BuildConfig(h))
+	if err == nil {
+		return client, nil
+	}
+	cs.Clients.Store(ns, c)
+	return c, nil
+}
+
+func (cs *ClientManager) GetClient(ns types.NamespacedName) (client ClientI, err error) {
```
dzeromski-hazelcast, line 34: If I understand it correctly we have signature like this: `client.ClientManager.GetClient()`. Can we remove `Client` prefix/suffix from struct/function names?

internal/hazelcast-client/status_service_manager.go
```diff
@@ -0,0 +1,44 @@
...
+
+type StatusServiceManager struct {
+	statusServices sync.Map
+}
+
+var (
+	errNoStatusService = errors.New("Status Service is not created yet")
+)
+
+func (ssm *StatusServiceManager) CreateStatusService(cl ClientI, l logr.Logger, ns types.NamespacedName, channel chan event.GenericEvent) *StatusService {
```
dzeromski-hazelcast, line 21: Do we need `Manager` for status service? Would it be enough to create service in reconciler with `ClientI` interface and hazelcast.Status reference as a parameter? Similar to what we do with HotBackup and `backup.Service`? Just fire and forget, let it run in background with `StatusServiceManager` etc.?

internal/hazelcast-client/backup_service.go
```diff
@@ -0,0 +1,46 @@
...
+	return &BackupService{
+		Client: cl,
+	}
+}
+
+func (bs *BackupService) ChangeClusterState(ctx context.Context, newState codecTypes.ClusterState) error {
+	req := codec.EncodeMCChangeClusterStateRequest(newState)
+	_, err := bs.Client.InvokeOnRandomTarget(ctx, req, nil)
+	if err != nil {
+		return fmt.Errorf("invoking: %w", err)
```
dzeromski-hazelcast, line 25: Do we add any information when we wrap error with `"invoking: %w"`? Can we just return `err`?

internal/hazelcast-client/client_manager.go
mtyazici: I already made the changes for that, it is now synchronous.

internal/hazelcast-client/client.go
mtyazici: Then we won't have a way of knowing if the operator is connected to the members when we report status.

internal/hazelcast-client/client.go
mtyazici: We can but as I said we won't have a way of reporting connection status in the operator.

internal/hazelcast-client/client.go
mtyazici: Same as above

internal/hazelcast-client/client.go
mtyazici: Not all requests are sent to all members, some of them sent to a single member. `InvokeOnMember` as you know returns `(*proto.ClientMessage, error)`, we would need to change at least the `ClientMessage` into a slice.

internal/hazelcast-client/client.go
mtyazici: Client is now synchronous as you said, but Hazelcast Go Client has a shutdown method that deletes all remaining resources. Should we not have a way of deleting them?

internal/hazelcast-client/client_manager.go
mtyazici: That is possible but for getter we should also have a parameter `h *hazelcastv1alpha1.Hazelcast` which is not needed in the `DeleteClient` function. If that is a fine approach for you we can do that.

internal/hazelcast-client/status_service_manager.go
mtyazici: Can you elaborate this, how are we going to access the status service in different reconcilers?

internal/hazelcast-client/backup_service.go
mtyazici: I agree we can just return the error

internal/hazelcast-client/client.go
mtyazici: done

dzeromski-hazelcast (approved):

internal/hazelcast-client/client.go
dzeromski-hazelcast: Ok, lets get back to this later.

internal/hazelcast-client/client.go
dzeromski-hazelcast: Ok, lets get back to this later.

internal/hazelcast-client/client.go
dzeromski-hazelcast: We could wrap in a slice *proto.ClientMessage or something. We can back to this later.

internal/hazelcast-client/status_service_manager.go
dzeromski-hazelcast: My mistake.

internal/hazelcast-client/config.go
```diff
@@ -18,6 +21,17 @@ func BuildConfig(h *hazelcastv1alpha1.Hazelcast) hazelcast.Config {
...
+			ConnectionStrategy: cluster.ConnectionStrategyConfig{
+				Timeout:       hztypes.Duration(3 * time.Second),
+				ReconnectMode: cluster.ReconnectModeOn,
+				Retry: cluster.ConnectionRetryConfig{
+					InitialBackoff: hztypes.Duration(200 * time.Millisecond),
+					MaxBackoff:     hztypes.Duration(1 * time.Second),
+					Jitter:         0.25,
+				},
+			},
+		},
```
cagric0, line 34: Why did we add this?

controllers/hazelcast/hazelcast_controller.go
```diff
@@ -220,14 +228,14 @@ func (r *HazelcastReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
...
 }
 
 func (r *HazelcastReconciler) runningPhaseWithStatus(req ctrl.Request) optionsBuilder {
-	if hzClient, ok := hzclient.GetClient(req.NamespacedName); ok {
-		return runningPhase().withStatus(hzClient.Status)
+	if ss, ok := r.statusServiceRegistry.Get(req.NamespacedName); ok {
+		return runningPhase().withStatus(ss.GetStatus())
 	}
 	return runningPhase()
 }
```
cagric0, line 241: I would move this method into `hazelcast_status.go` since this is status related method and now we also have another `HazelcastReconciler's` method (`update`) in that file.

cagric0 (approved):
LGTM, except one minor change.

internal/hazelcast-client/config.go
mtyazici: We made the client creation call blocking, so we needed to change the connection strategy. It also made sense to keep the config in one file.

mtyazici:
The failed test is known `[10.104.1.72]:5701 [dev] [5.2.1] Hot Restart procedure failed com.hazelcast.hotrestart.HotRestartException: Cluster-wide start failed!`

PR #401, opened 2022-10-17
[CN-499] Add support for restoring from HotBackup resource name

mtyazici:
Changes:
- Agent upload status now additionally returns a `message` and `backupKey`. `backupKey` is the `tgz` file the agent uploaded to bucket.
-  Updated `uploadOptions`, replaced `BackupFolderPath` with `BackupBaseDir` and `MemberUUID` with `MemberID`.
- Removed `backupType` from `persistence` field because it became obsolete.
- Exposed `hotBackupResourceName` in `persistence` field and implemented the necessary changes in HotBackup and Hazelcast controllers.

Agent side counter-part PR: https://github.com/hazelcast/hazelcast-platform-operator-agent/pull/32

PR #426, opened 2022-11-03
Bump 'platform-operator-agent' image version to 0.1.8

devOpsHelm:


hasancelik:
Agent version is already updated with https://github.com/hazelcast/hazelcast-platform-operator/pull/401

PR #428, opened 2022-11-03
[CN-577] Add Queue CR support

SeriyBg:


mtyazici (approved):
LGTM, maybe we can remove the changes under `config/crd/patches`.

bundle.yaml
```diff
@@ -2833,6 +2833,128 @@ status:
...
+    controller-gen.kubebuilder.io/version: v0.4.1
+  creationTimestamp: null
+  name: queues.hazelcast.com
+spec:
+  group: hazelcast.com
+  names:
+    kind: Queue
+    listKind: QueueList
+    plural: queues
+    singular: queue
```
hasancelik, line 2847: Should we provide shortName to users? I could not find the proper abbreviation for `queue`. `q` is the best one but it might conflict with other resources at the user's env.

hasancelik (approved):
LGTM 👍 with very minor comment 🙂

PR #433, opened 2022-11-09
[CN-596] Use struct definitions from hazelcast-platform-agent

dzeromski-hazelcast:
In order to use latest `hazelcast-platform-agent` I needed to fix few issues:

1. Needed to force golang version in github workers to avoid running on machines with old version: 
```
- uses: actions/setup-go@v3
  with:
    go-version: ">=1.19.0"
```
1. New golang version also required update of `sigs.k8s.io/controller-tools/cmd/controller-gen@v0.10.0` to be able to parse new golang code (any keyword, generics etc).
1. Update of controller-gen required regenerating operator yamls and bundles.


mtyazici (commented):
Should we also update agent version in `hazelcast_types.go` ?

dzeromski-hazelcast:
> Should we also update agent version in `hazelcast_types.go` ?

I think we can do it later. No change other than package was made.

mtyazici (approved):

hasancelik (approved):

PR #434, opened 2022-11-10
[CN-595] Add retry for backup service calls

mtyazici:


internal/backup/backup.go
```diff
@@ -49,15 +58,22 @@ func NewClusterBackup(h *hazelcastv1alpha1.Hazelcast) (*ClusterBackup, error) {
 
 func (b *ClusterBackup) Start(ctx context.Context) error {
 	// switch cluster to passive for the time of hot backup
-	err := b.client.ChangeClusterState(ctx, codecTypes.ClusterStatePassive)
+	err := retryOnError(backupBackoff,
+		func() error { return b.client.ChangeClusterState(ctx, codecTypes.ClusterStatePassive) })
```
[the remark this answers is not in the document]
mtyazici: What do you mean, which function signatures are the same?

internal/backup/backup.go
```diff
@@ -49,15 +58,22 @@ func NewClusterBackup(h *hazelcastv1alpha1.Hazelcast) (*ClusterBackup, error) {
...
+	defer retryOnError(backupBackoff,
+		func() error { return b.client.ChangeClusterState(ctx, codecTypes.ClusterStateActive) }) //nolint:errcheck })
+
+	return retryOnError(backupBackoff,
+		func() error { return b.client.TriggerHotRestartBackup(ctx) })
+}
 
-	return b.client.TriggerHotRestartBackup(ctx)
+func retryOnError(backoff wait.Backoff, fn func() error) error {
+	return retry.OnError(backoff, func(error) bool { return true }, fn)
```
dzeromski-hazelcast, line 76: Can we limit the errors somehow? Is there any sentinel error in the client that we could catch? I would like to avoid retrying if there is a network error like: connection refused and retry only on cluster state change errors.

dzeromski-hazelcast (approved):

internal/backup/backup.go
mtyazici: I think client can return `NonRetryableError` which is not very useful for deciding whether we should retry or not because we want to retry non-retryable functions anyway.

mtyazici: This is what it can return
```
func (s *Service) SendRequest(ctx context.Context, inv Invocation) error {
	select {
	case <-ctx.Done():
		return fmt.Errorf("sending invocation: %w", ctx.Err())
	case <-s.doneCh:
		return cb.WrapNonRetryableError(fmt.Errorf("sending invocation: %w", hzerrors.ErrClientNotActive))
	case s.requestCh <- inv:
		return nil
	}
}
```

internal/backup/backup.go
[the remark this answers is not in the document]
dzeromski-hazelcast: I see.

cagric0 (approved):

PR #381, opened 2022-10-05
[CN-545] Replace Rest client with Go for backup and refactor package structure

mtyazici:

PR #444, opened 2022-11-15
[CN-563] Add unit tests for HotBackup Reconciler

mtyazici:
Created unit tests are taken from the deleted tests in  #381
Also added unit test for the case where one member's timed member state returns failure, this case is created for the fixed issue #445

### DIFF
--- a/controllers/hazelcast/hot_backup_controller_test.go
+++ b/controllers/hazelcast/hot_backup_controller_test.go
@@ -157,7 +157,7 @@ func TestHotBackupReconciler_shouldNotTriggerHotBackupTwice(t *testing.T) {
 	Expect(hotBackupTriggers).Should(Equal(int32(1)))
 }
 
-func TestHotBackupReconciler_shouldCancelContextIfHazelcastCRIsDeleted(t *testing.T) {
+func TestHotBackupReconciler_shouldCancelContextIfHotbackupCRIsDeleted(t *testing.T) {
 	RegisterFailHandler(fail(t))
 	nn, h, hb := defaultCRs()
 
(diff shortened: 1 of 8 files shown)


controllers/hazelcast/hot_backup_controller_test.go
```diff
@@ -13,31 +23,181 @@ import (
...
+		defer reconcileWg.Done()
+		_, _ = r.Reconcile(context.TODO(), reconcile.Request{NamespacedName: nn})
+	}()
+	restCallWg.Done()
+	reconcileWg.Wait()
+
+	Expect(hotBackupTriggers).Should(Equal(int32(1)))
+}
+
+func TestHotBackupReconciler_shouldCancelContextIfHazelcastCRIsDeleted(t *testing.T) {
```
cagric0, line 160: Shouldn't this be .....If**HotbackupCR**IsDeleted?

cagric0 (commented):
I've made a little refactoring. If you are not okay with it, I can revert it. Thus, just one naming issue other than that LGTM!

controllers/hazelcast/hot_backup_controller_test.go
mtyazici:
```suggestion
func TestHotBackupReconciler_shouldCancelContextIfHotbackupCRIsDeleted(t *testing.T) {
```

You are right

cagric0 (approved):

SeriyBg (approved):

mtyazici:
Merged the PR since failure is unrelated to the PR.

PR #446, opened 2022-11-15
Fix error hiding in startBackup function

mtyazici:
startBackup function was returning no errors when one of the members failed with an error if the return value of the cancel function was nil. This resulted in backup process to continue with the next steps


mtyazici:
Closing the PR since the fix is already in main with 570894e49cfd23543299ccd2ec832d1273687eb8 and the tests will be added in #444

PR #450, opened 2022-11-16
[CN-590] public IP checks for ExposeExternally tests

semihbkgr:


semihbkgr:
As this PR is sent from the fork, tests get failed.
#459 It is the successor of this PR.

PR #455, opened 2022-11-21
 [CN-614] [5.5] Add additional port to the LB discovery service

mtyazici:
Backport of #453


SeriyBg (approved):

mtyazici:
> Is it possible to add a test case for this issue?

Problem occurs only on EKS and it is pretty hard to replicate in E2E tests since it depends on the environment. I would prefer we try to add  test when we have a proper solution to it.

PR #456, opened 2022-11-21
[CN-608] fix empty pod name field in hz member status

semihbkgr:


controllers/hazelcast/hazelcast_status.go
```diff
@@ -176,3 +185,17 @@ func (r *HazelcastReconciler) update(ctx context.Context, h *hazelcastv1alpha1.H
...
 	return ctrl.Result{}, nil
 }
+
+func getHzMemberPods(ctx context.Context, c client.Client, h *hazelcastv1alpha1.Hazelcast) []corev1.Pod {
+	podList := &corev1.PodList{}
+	namespace := client.InNamespace(h.Namespace)
+	matchingLabels := client.MatchingLabels{}
+	for k, v := range labels(h) {
+		matchingLabels[k] = v
+	}
```
mtyazici, line 195:
```suggestion
	matchingLabels := client.MatchingLabels(labels(h))
```

controllers/hazelcast/hazelcast_status.go
```diff
@@ -77,12 +79,19 @@ func (o optionsBuilder) withExternalAddresses(addrs string) optionsBuilder {
...
 	members := make([]hazelcastv1alpha1.HazelcastMemberStatus, 0, len(m))
 	for uid, member := range m {
 		a := member.Address
 		ip := a[:strings.IndexByte(a, ':')]
+		podName := ""
+		for _, pod := range memberPods {
+			if pod.Status.PodIP == ip {
+				podName = pod.Name
+			}
+		}
```
cagric0, line 92: It's not a big issue but, it's nice to prevent double loop. Instead of iterating over `memberPods` in each outer loop,  we could create a map with the key: `pod.Status.PodIP` value: `pod.Name` from `memberPods` before the outer loop. Then use that map in each loop to check if current `ip` exist in the map as a key. If so, use the value as the `podName`.

cagric0 (approved):
I left a tiny improvement comment, other than that LGTM!

mtyazici (approved):

PR #459, opened 2022-11-22
[CN-590] check public IPs in ExposeExternally tests

semihbkgr:


test/e2e/hazelcast_expose_externally_test.go
```diff
@@ -53,24 +59,133 @@ var _ = Describe("Hazelcast CR with expose externally feature", Label("hz_expose
...
+		By("matching HZ members with client members and comparing their public IPs")
+
+	memberLoop:
+		for _, member := range members {
+			for _, clientMember := range clientMembers {
+				if member.Uid == clientMember.UUID.String() {
+					service := getServiceOfMember(ctx, member)
+					Expect(service.Spec.Type).Should(Equal(corev1.ServiceTypeLoadBalancer))
+					Expect(service.Status.LoadBalancer.Ingress).Should(HaveLen(1))
+					serviceExternalIP := service.Status.LoadBalancer.Ingress[0].IP
```
mtyazici, line 131: This won't work at AWS, they use the hostname instead of IP, I think we should change it to work with AWS.

test/e2e/hazelcast_expose_externally_test.go
```diff
@@ -53,24 +59,133 @@ var _ = Describe("Hazelcast CR with expose externally feature", Label("hz_expose
...
+		clientHz := GetHzClient(ctx, hzLookupKey, false)
+		defer Expect(clientHz.Shutdown(ctx)).To(BeNil())
+		clientMembers := hzClient.NewClientInternal(clientHz).OrderedMembers()
+
+		By("matching HZ members with client members and comparing their public IPs")
+
+	memberLoop:
+		for _, member := range members {
+			for _, clientMember := range clientMembers {
+				if member.Uid == clientMember.UUID.String() {
```
mtyazici, line 86: Not too important but we try to [keep the happy path to left](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88). So, here you can reverse the if condition and move the logic to one tab left.
Code should look like:
```
if member.Uid != clientMember.UUID.String() {
  continue
}
// logic here
```

test/e2e/hazelcast_expose_externally_test.go
```diff
@@ -53,24 +59,133 @@ var _ = Describe("Hazelcast CR with expose externally feature", Label("hz_expose
...
```
mtyazici, line 127: Same as above

mtyazici (commented):
LGTM mostly, suggested a few changes.

cagric0 (approved):
LGTM except import reordering as in the comment

PR #461, opened 2022-11-23
[CN-591] Use health/ready endpoint in readinessProbe

dzeromski-hazelcast:


SeriyBg (approved):

controllers/hazelcast/hazelcast_resources.go
```diff
@@ -1014,29 +1016,34 @@ func (r *HazelcastReconciler) reconcileStatefulset(ctx context.Context, h *hazel
...
 						},
 					}},
-					TerminationGracePeriodSeconds: &[]int64{600}[0],
+					TerminationGracePeriodSeconds: pointer.Int64Ptr(600),
 				},
 			},
 		},
 	}
 
+	// for older versions fallback to health/node-stat endpoint
```
mtyazici, line 1035: Can we move this part into CreateOrUpdate function?

dzeromski-hazelcast: Just curious what is the rationale? We are not doing it for persistence etc. Should we move that too?

mtyazici: We are putting updatable things into createOrUpdate function, otherwise they will be set to the value they were created with.

hasancelik (commented):
Should we use `ready` endpoint for persistence disabled EE cluster and OS? Same question is valid for charts I think:
https://github.com/hazelcast/charts/pull/322#discussion_r1010156066
cc: @SeriyBg

mtyazici (approved):

SeriyBg:
@hasancelik @dzeromski-hazelcast, the `heath/ready` endpoint can be used when the persistence is disabled, as well as when for the `OS` version of Hazelcast. However, shouldn't we check for the version to be higher than `5.2`?

hasancelik (approved):

PR #464, opened 2022-11-23
Remove /main path from the tutorial link

cheels:
The link doesn't work with /main path in the end


cheels:
Duplicated here https://github.com/hazelcast/hazelcast-platform-operator/pull/470

PR #465, opened 2022-11-24
[CN-619] fix PR actions which is sent from a fork

semihbkgr:


.github/workflows/pull-request.yml
```diff
@@ -25,9 +25,31 @@ jobs:
   linter:
     name: Run linters
     runs-on: ubuntu-latest
+    permissions:
```
mtyazici, line 28: I think we should give permissions in workflow level instead of job level.

PR #478, opened 2022-11-29
[CN-621] E2E slow tests on GKE with Istio

semihbkgr:
## Description

GKE based E2E slow test suite with Istio enabled


.github/workflows/istio-e2e-kind.yaml
```diff
@@ -0,0 +1,190 @@
...
+
+      - name: Run Hazelcast E2E Tests
+        id: e2e-test
+        run: |
+          case ${{ matrix.edition }} in
+          os) GO_TEST_FLAGS=-ee=false;;
+          ee) GO_TEST_FLAGS=-ee=true;;
+          *)  echo Unexpected edition: ${{ matrix.edition }} && exit 1;;
+          esac
+          make test-e2e GO_TEST_FLAGS=${GO_TEST_FLAGS} NAMESPACE=$NAMESPACE NAME_PREFIX=$NAME_PREFIX WORKFLOW_ID=istio E2E_TEST_LABELS='&& !hz_expose_externally'
```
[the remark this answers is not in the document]
semihbkgr: Yes, `hz_expose_externally` is intentionally disabled, because current implementation is not supporting exposeExternally with Istio.

.github/workflows/istio-e2e-kind.yaml
```diff
@@ -0,0 +1,190 @@
+name: Istio E2E tests on Kind
+
+on:
+  pull_request:
+#  workflow_dispatch:
```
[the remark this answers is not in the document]
semihbkgr: I forgot to remove it.

.github/workflows/istio-e2e-kind.yaml
```diff
@@ -0,0 +1,193 @@
+name: Istio E2E tests on Kind
+
+on:
+  workflow_dispatch:
+    schedule:
+      - cron: "0 0 */3 * *"
```
cheels, line 6:
```suggestion
  workflow_dispatch:
  schedule:
      - cron: "0 0 */3 * *"
```

cheels (changes requested):
This is only one small issue in yaml should be fixed and then this is good to go...

.github/workflows/istio-e2e-kind.yaml
cheels, line 6: Also, why was chosen different from another workflow? Why it's not `- cron: "0 9 * * *"`

semihbkgr: As far as I know, this workflow should run once in three days.

cheels: I see, so  since it's a UTC, then it's better to use the same time as we use for Backup Workflow ("0 9 */3 * *")
It will run not at night but at the same time as other jobs but once in three days.

cheels (approved):

semihbkgr:
I add `E2E_TEST_EXCLUDED_SUITE` flag in makefile. It let you exclude explicit test suites easily in e2e tests

cheels (approved):

Makefile
```diff
@@ -176,11 +176,17 @@ E2E_TEST_SUITE ?= hz || mc || hz_persistence || hz_expose_externally || map || m
 ifeq (,$(E2E_TEST_SUITE))
 E2E_TEST_LABELS =
 else 
-E2E_TEST_LABELS = && $(E2E_TEST_SUITE)
+E2E_TEST_LABELS = && ($(E2E_TEST_SUITE))
+endif
+E2E_TEST_EXCLUDED_SUITE =
```
mtyazici, line 181: Does this not override what you give to the Makefile?

semihbkgr: What you mean?

mtyazici: It does not override it now I tested it :+1:

mtyazici (approved):

.github/workflows/istio-e2e-gke.yaml
```diff
@@ -0,0 +1,223 @@
...
+    secrets: inherit
+    with:
+      cluster_name: ${{ needs.create-gke-cluster.outputs.CLUSTER_NAME }}
+      cluster_type: gke
+      gh_run_id: ${{ github.run_id }}
+      gh_run_number: ${{ github.run_number }}
+      gh_sha: ${{ github.sha }}
+      namespaces: "test-operator-os, test-operator-ee"
+
+  get-image:
```
cheels, line 50: @semihbkgr Can you please change the step of getting the image? On the same way like it was done for GKE for example:
https://github.com/hazelcast/hazelcast-platform-operator/blob/main/.github/workflows/e2e-gke.yaml#L52-L90

And don't forget to use this everywhere where it needed https://github.com/hazelcast/hazelcast-platform-operator/blob/main/.github/workflows/e2e-gke.yaml#L104

cheels, line 50: @semihbkgr Can you please change the step of getting the image on the same way as it was done for GKE, for example:
https://github.com/hazelcast/hazelcast-platform-operator/blob/main/.github/workflows/e2e-gke.yaml#L52-L90

And don't forget to use this everywhere where it is needed https://github.com/hazelcast/hazelcast-platform-operator/blob/main/.github/workflows/e2e-gke.yaml#L104

cheels (changes requested):
Please update get_image step on the latest way

cheels (approved):

PR #479, opened 2022-11-29
ci: Update restart machines method for Open Shift environment

cheels:
## Description
The old method which is restarting failed nodes doesn't work. The current method checks if the machine set has replicas, if not it will be scaled to 0 and then scale up again to 1. Afterward, it's checked if all machines (**not machine set**) are in ready status. 

## User Impact

OCP nightly tests will be more stable and shouldn't fail if some of the nodes is not ready.

### DIFF
--- a/.github/scripts/utils.sh
+++ b/.github/scripts/utils.sh
@@ -158,29 +158,36 @@ cleanup_page_publish_runs()
             sleep 20
     done
 }
-# This function will restart all nodes that are not in ready status and wait until it will be ready
-wait_for_node_restarted()
+# This function will restart all machines that are not in ready status and wait until it will be ready
+wait_for_machine_restarted()
 {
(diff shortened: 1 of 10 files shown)


.github/scripts/utils.sh
```diff
@@ -158,31 +158,40 @@ cleanup_page_publish_runs()
             sleep 20
     done
 }
-# This function will restart all nodes that are not in ready status and wait until it will be ready
-wait_for_node_restarted()
+# This function will restart all machines that are not in ready status and wait until it will be ready
+wait_for_machine_restarted()
 {
    local TIMEOUT_IN_MINS=$1
    local NOF_RETRIES=$(( $TIMEOUT_IN_MINS * 3 ))
```
mtyazici, line 165: Just a note, you should tag the newly created EC2 nodes or the AWS node killer scripts will kill our nodes

PR #485, opened 2022-11-30
[CN-561] Add security context to all containers for hostPath permissions

mtyazici:
## Description

PR also contains addition of some field default values into containers because they triggered an update to statefulset every time the reconciler got triggered due to fields being overridden by the operator.

### DIFF
--- a/controllers/hazelcast/hazelcast_resources.go
+++ b/controllers/hazelcast/hazelcast_resources.go
@@ -1625,11 +1625,13 @@ func labels(h *hazelcastv1alpha1.Hazelcast) map[string]string {
 }
 
 func statefulSetAnnotations(h *hazelcastv1alpha1.Hazelcast) map[string]string {
-	var ans map[string]string
-	if h.Spec.ExposeExternally.IsSmart() {
-		ans[n.ServicePerPodCountAnnotation] = strconv.Itoa(int(*h.Spec.ClusterSize))
+	if !h.Spec.ExposeExternally.IsSmart() {
+		return nil
 	}
-	return ans
+
+	return map[string]string{
+		n.ServicePerPodCountAnnotation: strconv.Itoa(int(*h.Spec.ClusterSize)),
+	}
 }
 
 func podAnnotations(annotations map[string]string, h *hazelcastv1alpha1.Hazelcast) (map[string]string, error) {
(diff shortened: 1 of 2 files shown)


controllers/hazelcast/hazelcast_resources.go
```diff
@@ -1220,9 +1204,41 @@ func backupAgentContainer(h *hazelcastv1alpha1.Hazelcast) v1.Container {
 				MountPath: n.MTLSCertPath,
 			},
 		},
+		SecurityContext: containerSecurityContext(h),
 	}
 }
 
+func containerSecurityContext(h *hazelcastv1alpha1.Hazelcast) *v1.SecurityContext {
+	sec := &v1.SecurityContext{
+		RunAsNonRoot:             &[]bool{true}[0],
```
dzeromski-hazelcast, line 1213: Could you use also refactor it to use pointer.Bool() functions?

controllers/hazelcast/hazelcast_resources.go
```diff
@@ -1220,9 +1204,41 @@ func backupAgentContainer(h *hazelcastv1alpha1.Hazelcast) v1.Container {
...
+	if !h.Spec.Persistence.UseHostPath() {
+		return sec
+	}
+
+	sec.RunAsNonRoot = &[]bool{false}[0]
+	sec.RunAsUser = &[]int64{0}[0]
+
+	if platform.GetType() == platform.OpenShift {
+		sec.Privileged = &[]bool{true}[0]
+		sec.AllowPrivilegeEscalation = &[]bool{true}[0]
```
dzeromski-hazelcast, line 1236: Why do we need to escalate privileges in all containers?

dzeromski-hazelcast (approved):

controllers/hazelcast/hazelcast_resources.go
mtyazici: HostPath does not work otherwise in OpenShift since all containers can all change the files.

controllers/hazelcast/hazelcast_resources.go
mtyazici: sure :+1:

kutluhanmetin (approved):

SeriyBg (approved):

PR #488, opened 2022-12-01
Use 8 parallel Ginkgo processes for E2E tests

mtyazici:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -158,7 +158,7 @@ E2E_TEST_LABELS =
 else 
 E2E_TEST_LABELS = && $(E2E_TEST_SUITE)
 endif
-GINKGO_PARALLEL_PROCESSES ?= 2
+GINKGO_PARALLEL_PROCESSES ?= 4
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -158,7 +158,7 @@ E2E_TEST_LABELS =
 else 
 E2E_TEST_LABELS = && $(E2E_TEST_SUITE)
 endif
-GINKGO_PARALLEL_PROCESSES ?= 2
+GINKGO_PARALLEL_PROCESSES ?= 8
```
dzeromski-hazelcast, line 161: Could we check it on kind nightly workflow? I'm a little worried that right now we could have to little resources in our runner there.

dzeromski-hazelcast (approved):

kutluhanmetin (approved):

Makefile
mtyazici: I checked it, there were two Management Center errors and a MetalLB errors. I don't think the MetalLB error is related to this PR but MC error might be. Let's wait more before merging it.

PR #492, opened 2022-12-01
[CN-602] Add existing CR validation to validating webhook

dzeromski-hazelcast:
## Description

Add validating webhook for Hazlecast cluster and Topic data structure CRs. 

## User Impact

This change allows operator to report issues with CRs before they are applied.

### DIFF
--- a/api/v1alpha1/map_validation.go
+++ b/api/v1alpha1/map_validation.go
@@ -1,0 +1,70 @@
+package v1alpha1
+
+import (
+	"errors"
+)
+
+func ValidateNotUpdatableFields(current *MapSpec, last *MapSpec) error {
+	if current.Name != last.Name {
+		return errors.New("name cannot be updated")
+	}
+	if *current.BackupCount != *last.BackupCount {
+		return errors.New("backupCount cannot be updated")
+	}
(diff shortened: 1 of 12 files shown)


mtyazici (approved):

api/v1alpha1/map_validation.go
```diff
@@ -0,0 +1,70 @@
+package v1alpha1
+
+import (
+	"errors"
+)
+
+func ValidateNotUpdatableFields(current *MapSpec, last *MapSpec) error {
```
SeriyBg, line 7: What about the validation for the other CRs?

dzeromski-hazelcast: We will add as soon as we have it. This PR just adds existing validation do webhook where applicable.

SeriyBg: We do have the validation for the `Cache CR` with persistence 😃

dzeromski-hazelcast: Sure we can add it in https://github.com/hazelcast/hazelcast-platform-operator/pull/499 where we add Webhooks for cache :)

SeriyBg (approved):

PR #494, opened 2022-12-05
[CN-614] Use ServicePerPodLabelName when ExposeExternally is set

dzeromski-hazelcast:
## Description

Fix for first member trying to assume external IP of the discovery service load balancer.


hasancelik (commented):
Should we write a test for it? Is it hard?

cagric0 (approved):

PR #495, opened 2022-12-05
[CN-637] fixing expose externally tests on EKS

semihbkgr:
## Description

It fixes the failing expose externally tests on EKS.

Fixed tests:

- `[It] should create Hazelcast cluster exposed with LoadBalancer services and allow connecting with Hazelcast smart client` 

The reason why this test gets failed is that the hostname of the service created by EKS has more than one public IP addresses.

## User Impact

### DIFF
--- a/test/e2e/hazelcast_expose_externally_test.go
+++ b/test/e2e/hazelcast_expose_externally_test.go
@@ -135,11 +135,22 @@ var _ = Describe("Hazelcast CR with expose externally feature", Label("hz_expose
 				service := getServiceOfMember(ctx, hzLookupKey.Namespace, member)
 				Expect(service.Spec.Type).Should(Equal(corev1.ServiceTypeLoadBalancer))
 				Expect(service.Status.LoadBalancer.Ingress).Should(HaveLen(1))
-				serviceExternalIP := getLoadBalancerIngressPublicIP(ctx, service.Status.LoadBalancer.Ingress[0])
+				svcLoadBalancerIngress := service.Status.LoadBalancer.Ingress[0]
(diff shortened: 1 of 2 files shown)


mtyazici (approved):

test/e2e/hazelcast_expose_externally_test.go
```diff
@@ -135,11 +135,22 @@ var _ = Describe("Hazelcast CR with expose externally feature", Label("hz_expose
...
 				Expect(clientPublicAddresses).Should(HaveLen(1))
 				clientPublicIp := clientPublicAddresses[0][:strings.IndexByte(clientPublicAddresses[0], ':')]
-				Expect(serviceExternalIP).Should(Equal(clientPublicIp))
+				if svcLoadBalancerIngress.IP != "" {
+					Expect(svcLoadBalancerIngress.IP).Should(Equal(clientPublicIp))
+				} else {
+					hostname := svcLoadBalancerIngress.Hostname
+					var addressMatched bool
+					Eventually(func() error {
+						matched, err := DnsLookupAddressMatched(ctx, hostname, clientPublicIp)
```
SeriyBg, line 148: I think we can inline this function inside the `Eventually` block, IMO it would be a bit more readable
```
Eventually(func() string, error {
	IPs, err := net.DefaultResolver.LookupHost(ctx, hostname)
	if err != nil {
		return "", err
	}
	for _, IP := range IPs {
		if IP == addr {
			return IP, nil
		}
	}
	return "", nil
}, 3*Minute, interval).Should(Equal(clientPublicIp))
```

PR #501, opened 2022-12-13
[CN-639] Fix Hazelcast ConfigMap when multiple CRs for Data Structures are created

SeriyBg:
## Description

When multiple Data Structures CR was created, only the last one was stored in the `hazelcast.yaml` ConfigMap

## User Impact

User can create multiple CR and all will be stored in the ConfigMap


kutluhanmetin (approved):
LGTM!

mtyazici (approved):

PR #503, opened 2022-12-15
[CN-643] Allow to create HotBackup before Hazelcast cluster is ready

dzeromski-hazelcast:
## Description

Do not return error when Hazelcast CR is not ready yet.

### DIFF
--- a/test/e2e/hazelcast_persistence_test.go
+++ b/test/e2e/hazelcast_persistence_test.go
@@ -11,6 +11,7 @@ import (
 	appsv1 "k8s.io/api/apps/v1"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 
+	hazelcastcomv1alpha1 "github.com/hazelcast/hazelcast-platform-operator/api/v1alpha1"
 	hazelcastv1alpha1 "github.com/hazelcast/hazelcast-platform-operator/api/v1alpha1"
 	codecTypes "github.com/hazelcast/hazelcast-platform-operator/internal/protocol/types"
 	"github.com/hazelcast/hazelcast-platform-operator/test"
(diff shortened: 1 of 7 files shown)


test/e2e/hazelcast_persistence_test.go
```diff
@@ -252,4 +252,25 @@ var _ = Describe("Hazelcast CR with Persistence feature enabled", Label("hz_pers
...
+		setLabelAndCRName("hp-8")
+
+		clusterSize := int32(1)
+		hazelcast := hazelcastconfig.HazelcastPersistencePVC(hzLookupKey, clusterSize, labels)
+		hazelcast.Spec.Persistence.ClusterDataRecoveryPolicy = hazelcastv1alpha1.MostRecent
+
+		By("creating HotBackup CR")
+		hotBackup := hazelcastconfig.HotBackup(hbLookupKey, hazelcast.Name, labels)
+		Expect(k8sClient.Create(context.Background(), hotBackup)).Should(Succeed())
+
```
SeriyBg, line 269: I would add an assert here for the `HotBackupNotStarted` status

SeriyBg (approved):
One small suggestion, but LGTM

test/e2e/hazelcast_persistence_test.go
dzeromski-hazelcast: I was thinking about it and decided no to do it. Do we care about this status? We just want backup to start and dont care what is the status before.

SeriyBg: Well, IMO we do care that the `HotBackup` is started even after being set to the `NotStarted` state. So this way will eliminate the rare but possible case when the HB reconcile is executed after the `Hazelcast CR` is created

dzeromski-hazelcast: Is it a bug? Running HB reconcile is executed after the Hazelcast CR is created shouldn't be a problem, right?

SeriyBg: It shouldn't be a problem, of course. It's just that I would prefer to have an assertion that proves that with this scenario our logic still works fine 😃

cagric0 (approved):

test/e2e/hazelcast_persistence_test.go
dzeromski-hazelcast: Fixed.

PR #515, opened 2022-12-22
[CN-649] Change nightly test triggers

mtyazici:
## Description

Now the tests are triggered when:
- `run-all-test` label is added into PR
- or when the PR is merged into master branch and `run-all-test` label is not present in the PR

Also removed nightly Kind test altogether since we have PR builders running on Kind.


cagric0 (approved):
LGTM!

.github/workflows/e2e-ocp.yaml
```diff
@@ -112,7 +110,7 @@ jobs:
       - name: Build Image
         if: github.event_name == 'workflow_dispatch'
```
cheels, line 111: I think the workflow_dispatch condition also should be removed

.github/workflows/e2e-ocp.yaml
```diff
@@ -89,7 +87,7 @@ jobs:
     env:
       IMG: hazelcast/hazelcast-platform-operator:latest-snapshot
```
cheels, line 88: Since we are building an image and pushing it to ttl.sh the latest snapshot doesn't require here

.github/workflows/k8s-dist-tests.yaml
```diff
@@ -0,0 +1,108 @@
...
+      - name: Decide to run tests
+        id: should-run-tests
+        run: |
+          if [[ '${{ github.event_name }}' != 'push' ]]; then
+            echo "RUN_TESTS=true" >> $GITHUB_OUTPUT
+            exit 0
+          fi
+
+          COMMIT_MESSAGE='${{ github.event.commits[0].message }}'
+          PR_NUMBER=$(echo $COMMIT_MESSAGE | grep -oP '\(#\K[0-9]*(?=\))')
```
cheels, line 69: Can we get the PR number using this, or it's required to get it from the COMMIT message?

${{ github.event.number }}

mtyazici: Is there an event number? Do you have a link to the documentation?

.github/workflows/e2e-ocp.yaml
mtyazici: I will delete it

.github/workflows/e2e-ocp.yaml
mtyazici: I will delete it

cheels (approved):

PR #519, opened 2022-12-23
[CN-652]-Add focused tests validator into PR builder

cheels:
## Description

Added focused test validation into PR builder. The PR builder immediately fails if one of the tests has an 'F' leftover.

## User Impact

No more focused tests in the codebase.

### DIFF
--- a/.github/workflows/pull-request.yml
+++ b/.github/workflows/pull-request.yml
@@ -20,6 +20,7 @@ permissions:
 
 env:
   GO_VERSION: '1.19.0'
+  GOBIN: ${{ github.workspace }}/bin
 
 concurrency:
   group: ${{ github.workflow }}-${{ github.ref }}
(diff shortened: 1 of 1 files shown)


.github/workflows/pull-request.yml
```diff
@@ -62,6 +63,15 @@ jobs:
         with:
           config_file: hack/yamllint.yaml
 
+      - name: Check if tests are unfocused
+        if: ${{ !cancelled() }}
+        run: |
+           go install -mod=mod github.com/onsi/ginkgo/v2/ginkgo
```
mtyazici, line 69: Why are we overriding GOBIN, is it not set? We could use `make ginkgo` command to install ginkgo.

cheels: @mtyazici Yes GOBIN is always empty. 
image

As for the make ginkgo - that is a good point. I'll change it 👍

mtyazici (approved):

hasancelik (approved):

PR #522, opened 2022-12-27
test: Add tests for K8s repartitioning functionality

cheels:
## Description
Added 2 tests that test repartitioning events, which shouldn't happen after the planned shutdown and during the temporary loss of members. In both cases repartitioning shouldn't happen.


cagric0 (approved):
LGTM! Perfect tests!

SeriyBg (approved):

PR #523, opened 2022-12-28
[CN-631]-add sync_tags function to fetch latest tag

cheels:
## Description

- Added sync_tag function that fetched and attached the latest tag to the published certificate image on Open Shift env
- Changed the tag for the ubi-minimal image to fix the vulnerability issue. Publishing with the current tag will get an index score of C instead of A, and it will cause the impossibility to publish our next image
image

## User Impact

- No more needs to sync tags manually after publishing
- From now we should always get an index score of A for our subsequent releases

image

### DIFF
--- a/.github/scripts/utils.sh
+++ b/.github/scripts/utils.sh
@@ -192,3 +192,20 @@ wait_for_instance_restarted()
       echo "All instances are in 'Ready' status."
    fi
 }
+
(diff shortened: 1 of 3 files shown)


mtyazici (approved):

kutluhanmetin (approved):
LGTM

PR #541, opened 2023-01-03
[CN-449] add connectivity check to wan replication cr

kutluhanmetin:
## Description

- While applying WAN Replication
```
cat <<EOF | kubectl apply -f -
kind: WanReplication
metadata:
  name: example-wan-replication
spec:
  resources:
    - name: hazelcast-first
      kind: Hazelcast
  targetClusterName: dev
  endpoints: "<SECOND-CLUSTER-EXTERNAL-IP>"
EOF
```

The sidecar container inside Hazelcast Pod tries a TCP Connection to the target cluster specified as `SECOND-CLUSTER-EXTERNAL-IP`. 

## User Impact
When creating a WAN Replication CR, user is going to be sure that target cluster is reachable.


kutluhanmetin:
This is blocked by: https://github.com/hazelcast/hazelcast-platform-operator-agent/pull/40. Before merging this PR, new agent version must be released, and it must be updated in operator with this PR.

controllers/hazelcast/wanreplication_controller.go
```diff
@@ -165,6 +190,30 @@ func (r *WanReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	return updateWanStatus(ctx, r.Client, wan, wanSuccessStatus())
 }
 
+func checkConnectivity(ctx context.Context, m *mtls.Client, endpoints string, memberAddresses []string) error {
+	for _, memberAddress := range memberAddresses {
```
dzeromski-hazelcast, line 194: I think we can run test on a random node. No need to check if all of them have access. What do you think?

controllers/hazelcast/wanreplication_controller.go
```diff
@@ -165,6 +190,30 @@ func (r *WanReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+		p, err := pinger.NewPinger(&pinger.Config{
+			MemberAddress: memberAddress,
+			MTLSClient:    m,
+		})
+		if err != nil {
+			return err
+		}
+
+		ping, err := p.Ping(ctx, endpoints)
```
dzeromski-hazelcast, line 203: Can we drop `ping` return value? Looks like `ping` and `err` are exclusive. Only one variable is set so we can simplify and just check for `err`?

controllers/hazelcast/wanreplication_controller.go
```diff
@@ -3,7 +3,10 @@ package hazelcast
 import (
 	"context"
 	"encoding/json"
+	"errors"
 	"fmt"
+	"github.com/hazelcast/hazelcast-platform-operator/internal/mtls"
```
dzeromski-hazelcast, line 8: can we keep stdlib imports separate from our and third-party imports?

internal/pinger/pinger.go
```diff
@@ -0,0 +1,44 @@
+package pinger
+
+import (
+	"context"
+	"github.com/hazelcast/hazelcast-platform-operator/internal/mtls"
+	"github.com/hazelcast/hazelcast-platform-operator/internal/rest"
```
dzeromski-hazelcast, line 6: Separate stdlib imports? :)

internal/rest/pinger_service.go
```diff
@@ -0,0 +1,47 @@
...
+		return nil, err
+	}
+
+	return &PingerService{client: &Client{
+		BaseURL: b,
+		client:  c,
+	}}, nil
+}
+
+type PingRequest struct {
```
dzeromski-hazelcast, line 25: Can we define `PingRequest` and `PingResponse` in operator-agent and then import it? Just like we do for other services?

internal/rest/pinger_service.go
```diff
@@ -0,0 +1,47 @@
...
+
+func (p *PingerService) Ping(ctx context.Context, r *PingRequest) (*PingResponse, *http.Response, error) {
+	u := "ping"
+
+	uploadReq, err := p.client.NewRequest("POST", u, r)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	var uploadResp *PingResponse
```
dzeromski-hazelcast, line 41: We define here `*PingResponse` nil pointer. Shouldn't we define normal struct `PingResponse` here and pass pointer to `.Do` methods so that we have a storage for the struct?

internal/pinger/pinger.go
```diff
@@ -0,0 +1,44 @@
...
+	service *rest.PingerService
+	config  *Config
+}
+
+type Config struct {
+	MemberAddress string
+	MTLSClient    *mtls.Client
+}
+
+func NewPinger(config *Config) (*Pinger, error) {
```
dzeromski-hazelcast, line 20: Not sure this is the best name. Could we maybe switch to something like "ConnectivityService" or "Dial". We are not doing icmp pings like the name "Pinger" could imply.

dzeromski-hazelcast (approved):

controllers/hazelcast/wanreplication_controller.go
kutluhanmetin: I didn't understand the reason actually. Because we have a source cluster and a target cluster. Target cluster should be reachable from all the members of the source cluster I think.

internal/rest/pinger_service.go
kutluhanmetin: Well, I think it is a general issue we need to talk about it. Because I didn't release the agent yet, It is also in review. I think we have two options to resolve this problem:
- Monorepo (I strongly suggest adding monorepo codebase to operator repository)
- Break the dependency to agent code like I did in this PR, and keep them sync all the time.

controllers/hazelcast/wanreplication_controller.go
kutluhanmetin: Done 👍

internal/pinger/pinger.go
kutluhanmetin: Done 👍

internal/rest/pinger_service.go
kutluhanmetin: Done 👍

controllers/hazelcast/wanreplication_controller.go
kutluhanmetin: Done 👍

api/v1alpha1/hazelcast_types.go
```diff
@@ -272,12 +272,12 @@ func (c *UserCodeDeploymentConfig) IsConfigMapEnabled() bool {
...
 type AgentConfiguration struct {
 	// Repository to pull Hazelcast Platform Operator Agent(https://github.com/hazelcast/platform-operator-agent)
-	// +kubebuilder:default:="docker.io/hazelcast/platform-operator-agent"
+	// +kubebuilder:default:="docker.io/kutluhanhazelcast/platform-operator-agent"
 	// +optional
 	Repository string `json:"repository,omitempty"`
 
 	// Version of Hazelcast Platform Operator Agent.
-	// +kubebuilder:default:="0.1.12"
+	// +kubebuilder:default:="0.1.14"
```
mtyazici, line 280: We should revert it when we decide to merge the PR.

controllers/hazelcast/wanreplication_controller.go
```diff
@@ -63,6 +69,45 @@ func (r *WanReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 	}
 	ctx = context.WithValue(ctx, LogKey("logger"), logger)
 
+	for _, rr := range wan.Spec.Resources {
```
mtyazici, line 72: Should we take this part into a function? Also this logic should be put after the finalizer executor logic since any error in this part can prevent the resource from getting deleted.

mtyazici (changes requested):
Left some comments about the reconciler function :+1:

api/v1alpha1/hazelcast_types.go
kutluhanmetin: I know 👍

internal/rest/pinger_service.go
kutluhanmetin: Added a retro item for this one

controllers/hazelcast/wanreplication_controller.go
```diff
@@ -165,6 +176,56 @@ func (r *WanReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Reque
...
+			}
+			hzResourceName = m.Spec.HazelcastResourceName
+		}
+
+		statusService, ok := r.statusServiceRegistry.Get(types.NamespacedName{
+			Namespace: req.Namespace,
+			Name:      hzResourceName,
+		})
+		if !ok {
+			logger.Error(errors.New("get Hazelcast Status Service failed"), "")
```
mtyazici, line 202: Should we return the error here instead of logging?

kutluhanmetin: Done.

controllers/hazelcast/wanreplication_controller.go
dzeromski-hazelcast: Does replication run on all nodes or just the main node? In addition iterating over all members will be slow for large clusters. If you want to keep it then we should use concurrency here like we do in backup and restore.

kutluhanmetin: Added the following commit on agent side to make it concurrent and limit with timeout:
https://github.com/hazelcast/hazelcast-platform-operator-agent/pull/40/commits/4d5de46bf8de16e09bc2baa034b3233470bfca99

mtyazici (approved):

PR #557, opened 2023-01-10
ci: Update test status badges in README.md file

cheels:
## Description

Fix status badges on the main repo page in the README.md file:
- The current solution fetches the statuses of the tests from the run id, generates a custom JSON response and puts it into the Gist under devOpsHelm user
- In case of Gist will be deleted suddenly, the job will recreate it and ask the user to update the secret and provide updated links for README.md file with the proper gist id for further use. 
- If impossible to fetch the data - the JSON file will be generated with unknown status and grey color.

## User Impact

Test status badges should display correct statuses.


mtyazici:
If we do not run the tests in the PR and run them in the main, what will the badges show? Will  it show the latest PR run instead of the main?

cheels:
> If we do not run the tests in the PR and run them in the main, what will the badges show? Will it show the latest PR run instead of the main?

Good catch. No it will not update the badges. I need to add this to each particular job with workflow_dispatch condition.

mtyazici (approved):

cagric0 (approved):

PR #566, opened 2023-01-16
[CN-664] Check if licenseKeySecret exists

dzeromski-hazelcast:
## Description

We now check if specified `licenseKeySecret` (secret name) actually exists in Kubernetes namespace.

## User Impact

Hazelcast cluster status will change to `Failed` with message `Secret Not Found` if user tries to use non existing licenseKey secret.


controllers/hazelcast/hazelcast_controller.go
```diff
@@ -117,6 +117,24 @@ func (r *HazelcastReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 				withMessage(fmt.Sprintf("error validating new Spec: %s", err)))
 	}
 
+	// make sure secret exists, we do it here because of easy
+	// access to kubernetes client (r.Client)
```
mtyazici, line 121: I agree it is easier this way but shouldn't we give the client to ValidateHazelcastSpec somehow, so we can do more sophisticated validation later. I know it is not straightforward to add the client for the webhook but if we could it would be easier for future validations.

mtyazici (approved):

internal/kubeclient/client.go
```diff
@@ -0,0 +1,49 @@
...
+
+import (
+	"context"
+
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/manager"
+)
+
+// DefaultClient is the default Kubernetes Client
+var DefaultClient client.Reader = &nopClient{}
```
SeriyBg, line 11: Do we need this var to be public?

SeriyBg (approved):

PR #567, opened 2023-01-16
[CN-671] Support -XX:{Initial,Max,Min}RAMPercentage Java options

dzeromski-hazelcast:
## Description

Following configuration parameters were added to Hazelcast CR:
1. spec.jvm.memory.initialRAMPercentage (`-XX:InitialRAMPercentage`)
2. spec.jvm.memory.minRAMPercentage (`-XX:MinRAMPercentage`)
3. spec.jvm.memory.maxRAMPercentage (`-XX:MaxRAMPercentage`)

In addition by default we add `-XX:+UseContainerSupport` to `JAVA_OPTS`.

## User Impact

User can now configure  `-XX:{Initial,Max,Min}RAMPercentage` JVM options using Hazelcast CR.

### DIFF
--- a/api/v1alpha1/hazelcast_types.go
+++ b/api/v1alpha1/hazelcast_types.go
@@ -132,6 +132,11 @@ type HazelcastSpec struct {
 	// +optional
 	// +kubebuilder:default:={}
 	HighAvailabilityMode HighAvailabilityMode `json:"highAvailabilityMode,omitempty"`
+
+	// Hazelcast JVM configuration
+	// +optional
+	// +kubebuilder:default:={}
+	JVM *JVMConfiguration `json:"jvm,omitempty"`
 }
 
 // +kubebuilder:validation:Enum=NODE;ZONE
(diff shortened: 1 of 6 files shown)


kutluhanmetin (approved):
LGTM

api/v1alpha1/hazelcast_types.go
```diff
@@ -560,6 +565,56 @@ func (c *ExposeExternallyConfiguration) MemberAccessServiceType() corev1.Service
...
+
+func (c *JVMConfiguration) GetMemory() *JVMMemoryConfiguration {
+	if c != nil {
+		return c.Memory
+	}
+	return nil
+}
+
+// JVMMemoryConfiguration is a JVM memory configuration
+type JVMMemoryConfiguration struct {
```
mtyazici, line 583: Just a note, if the empty string value is not valid for these fields we can simplify the getters by removing the pointers for all the fields.

dzeromski-hazelcast: In this case they are optional so it makes sense.

But you are right in pointing out "getters". I'm experimenting with this new approach to API. I'm not convinced about it yet but this is how official protobuf go library is doing it. Lets keep it and see if works?

mtyazici: Sure :+1:

dzeromski-hazelcast:
> Could you add a sample YAML for this feature?

Done.

hasancelik (approved):

PR #569, opened 2023-01-16
[CN-701] reunite client registry get and create methods

kutluhanmetin:
## Description

With this fix, we will create clients for Hazelcast CRs that marked as created.


controllers/hazelcast/map_controller.go
```diff
@@ -209,9 +209,9 @@ func ValidatePersistence(pe bool, h *hazelcastv1alpha1.Hazelcast) error {
 	return nil
 }
 
-func GetHazelcastClient(cs hzclient.ClientRegistry, m *hazelcastv1alpha1.Map) (hzclient.Client, error) {
-	hzcl, ok := cs.Get(types.NamespacedName{Name: m.Spec.HazelcastResourceName, Namespace: m.Namespace})
-	if !ok {
+func GetHazelcastClient(cs hzclient.Registry, m *hazelcastv1alpha1.Map) (hzclient.Client, error) {
+	hzcl, err := cs.GetOrCreate(context.Background(), types.NamespacedName{Name: m.Spec.HazelcastResourceName, Namespace: m.Namespace})
```
SeriyBg, line 213: Can we pass the `ctx` to the `GetHazelcastClient` and reuse it instead of creating a new one?

SeriyBg (approved):
Just one small comment. Other than that LGTM

mtyazici:
@kutluhanmetin,

We might also need status registry to be changed, AFAIR, you added status registry to wan replication. So it would be better to do the same changes for `statusRegistry` interface.

Just a note also, I think changing `ClientRegistry` to `Registry` is a bit confusing since in the same package we have `StatusRegistry`. Maybe we should change the name of the package, wdyt?

internal/util/util.go
```diff
@@ -28,7 +28,7 @@ func CreateOrUpdate(ctx context.Context, c client.Client, obj client.Object, f c
 	opResult, err := controllerutil.CreateOrUpdate(ctx, c, obj, f)
 	if kerrors.IsAlreadyExists(err) {
 		// Ignore "already exists" error.
-		// Inside createOrUpdate() there's is a race condition between Get() and Create(), so this error is expected from time to time.
+		// Inside createOrUpdate() there's is a race condition between Get() and GetOrCreate(), so this error is expected from time to time.
```
mtyazici, line 31: I think this should stay the same

mtyazici:
Are we going to write tests in this PR for the operator deletion case? If we do, we might want to add it in this PR.

controllers/hazelcast/map_controller.go
kutluhanmetin: Done

mtyazici (approved):

PR #570, opened 2023-01-16
[CN-672] Add Native Memory Configuration

dzeromski-hazelcast:
## Description

Native Memory can be configured in Hazelcast CR:
```
kind: Hazelcast
metadata:
  name: <name>
spec:
  nativeMemory:
    enabled: true | false
    allocatorType: STANDARD | POOLED
    size: <quantity>
    minBlockSize: <numeric_value>
    pageSize: <numeric_value>
    metaDataSpacePercentage: <numeric_value>
```
We allow `nativeMemory.min-block-size`, `nativeMemory.page-size`, `nativeMemory.metadata-space-percentage` properties only if `nativeMemory.allocatorType` is set to `POOLED`.

## User Impact

User is now able to specify HD Memory configuration using Hazelcast CR


mtyazici:
Just glanced over the api changes, is there a reason we do not give information about the default fields to the users? We are wrapping the Hazelcast configuration fields and users might not be aware that the API we expose have default values.

api/v1alpha1/hazelcast_types.go
```diff
@@ -560,6 +565,91 @@ func (c *ExposeExternallyConfiguration) MemberAccessServiceType() corev1.Service
...
+	PageSize *int32 `json:"pageSize,omitempty"`
+
+	// MetadataSpacePercentage defines percentage of the allocated native memory
+	// that is used for the metadata of other map components such as index
+	// (for predicates), offset, etc.
+	// +optional
+	MetadataSpacePercentage *int32 `json:"metadataSpacePercentage,omitempty"`
+}
+
+func (c *NativeMemoryConfiguration) GetEnabled() bool {
```
kutluhanmetin, line 611: Would you remove `Get` prefixes for these methods?

kutluhanmetin (approved):

api/v1alpha1/hazelcast_types.go
dzeromski-hazelcast: Unfortunately not. We cannot have the same field and method names:
```go
type Foo struct {
	Bar struct{}
}

func (f *Foo) Bar() {}
```
https://go.dev/play/p/r3inF92JeFN

Will give us an error:
```
./prog.go:7:15: field and method with the same name Bar
	./prog.go:4:2: other declaration of Bar
```

api/v1alpha1/hazelcast_validation.go
```diff
@@ -91,3 +95,29 @@ func checkEnterprise(repo string) bool {
 	}
 	return strings.HasSuffix(path[len(path)-1], "-enterprise")
 }
+
+func validateNativeMemory(h *Hazelcast) error {
```
hasancelik, line 99: > We allow nativeMemory.min-block-size, nativeMemory.page-size, nativeMemory.metadata-space-percentage properties only if nativeMemory.allocatorType is set to POOLED

Could you write an integration tests for the scenarios below?

hasancelik (changes requested):

api/v1alpha1/hazelcast_types.go
```diff
@@ -560,6 +565,91 @@ func (c *ExposeExternallyConfiguration) MemberAccessServiceType() corev1.Service
...
+
+	// NativeMemoryPooled is Hazelcast own pooling memory allocator
+	NativeMemoryPooled NativeMemoryAllocatorType = "POOLED"
+)
+
+// NativeMemoryConfiguration is a Hazelcast HD memory configuration
+type NativeMemoryConfiguration struct {
+	// Enabled specifies if HD Memory is enabled
+	// +optional
+	Enabled *bool `json:"enabled,omitempty"`
```
hasancelik, line 584: I think we do not need separate `enabled` parameter 🙂

hasancelik (approved):

PR #582, opened 2023-01-20
[CN-673] Support Native Memory format in Map CR

dzeromski-hazelcast:
## Description

Allow using Off-Heap Memory (HD Memory, Native Memory) in Map CR with spec.inMemoryFormat: NATIVE configuration.


kutluhanmetin (approved):

mtyazici (approved):
I think we should also update the Map CR documentation. 

My other question is, are we going to create a doc page for native memory support in general?

PR #583, opened 2023-01-22
[CN-687] Custom Logger support for the go client

semihbkgr:
## Description

Implementing Hazelcast client custom logger adapter for `logr` library which is used in the operator.

For now, it is only for debug-purpose. So there is no config field to set the level of Hz client logger. The level is set to `logger.InfoLevel` in hard-coded way.

### DIFF
--- a/internal/hazelcast-client/logger.go
+++ b/internal/hazelcast-client/logger.go
@@ -1,0 +1,44 @@
+package client
+
+import (
+	"fmt"
+	"github.com/go-logr/logr"
+	"github.com/hazelcast/hazelcast-go-client/logger"
+	hazelcastv1alpha1 "github.com/hazelcast/hazelcast-platform-operator/api/v1alpha1"
+	"github.com/hazelcast/hazelcast-platform-operator/internal/util"
+)
+
+type LogrHzClientLoggerAdapter struct {
+	logger logr.Logger
+	weight logger.Weight
+}
+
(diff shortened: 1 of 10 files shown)


internal/hazelcast-client/logger.go
```diff
@@ -0,0 +1,51 @@
...
+
+type LogrHzClientLoggerAdapter struct {
+	logger     logr.Logger
+	weight     logger.Weight
+	enableFunc func() bool
+}
+
+func NewLogrHzClientLoggerAdapter(rootLogger logr.Logger, level logger.Level, h *hazelcastv1alpha1.Hazelcast) (*LogrHzClientLoggerAdapter, error) {
+	w, err := logger.WeightForLogLevel(level)
+	l := rootLogger.WithName("client").WithName(h.Name).
```
mtyazici, line 18: rootLogger's value might differ when client is created from different reconcilers, do we have a way of reverting that value?

semihbkgr: `rootLogger` does not have any value in the current use case. It is from `ctrl.Log` controller-runtime logger. In the future, we may want to pass more specific root logger with its values.

mtyazici, line 18: We use `h.Name` two times, should drop one of them.

mtyazici (approved):
LGTM :+1:

internal/hazelcast-client/logger.go
semihbkgr: `.WithName(h.Name)` to make client logger name unique
`WithValues("Hazelcast", h.Name)` to enable filtering HZ CR logs with their client logs

hasancelik (approved):

PR #585, opened 2023-01-23
[CN-699] Add MTLS secret generation by namespace

SeriyBg:
## User Impact

Support MTLS connection to Sidecar per namespace


internal/mtls/client_registry.go
```diff
@@ -0,0 +1,52 @@
...
+
+func NewHttpClientRegistry() HttpClientRegistry {
+	return &httpClientRegistry{}
+}
+
+type httpClientRegistry struct {
+	clients sync.Map
+}
+
+func (cr *httpClientRegistry) Create(ctx context.Context, kubeClient client.Client, ns string) (*http.Client, error) {
```
kutluhanmetin, line 28: Should we make it GetOrCreate as we did in Hazelcast Client Registry?

kutluhanmetin (approved):

internal/mtls/client_registry.go
SeriyBg: Well,here is a bit different, IMO, as if the `Hazelcast CR` did not create the secret, then there is no need for it in other reconcilers

PR #589, opened 2023-01-25
[CN-698] Add cluster scoped operator support to Helm chart

mtyazici:
## Description

Changes:
- Operator by default watches all namespaces.
- Operator creates RBAC resources in the following three main cases
  - Watching Own Namespace: Add watched namespace rules to Role. If node discovery is enabled at the same time, create ClusterRole/Binding for that.
  - Watching Other Namespace: Add watched namespace rules to Role and use RoleBinding which are created in watched namespace. If node discovery is enabled at the same time also create ClusterRole/Binding for that as well.
  - Watching All Namespace: Add watched namespace rules to ClusterRole and use ClusterroleBinding. If node discovery is enabled use the same ClusterRole/Binding.

## User Impact

Operator can now watch all namespaces by configuring `watchedNamespace` value in Helm Chart.


helm-charts/hazelcast-platform-operator/values.yaml
```diff
@@ -16,11 +16,12 @@ phoneHomeEnabled: true
...
 # If false and Helm is misconfigured, it will fail the release.
-allowClusterScopedPermissions: true
+createClusterScopedResources: true
 webhook:
   enabled: true
-# Namespace that operator watches. If empty string or omitted, release namespace is watched.
-watchedNamespace: ""
+# Namespace that operator watches.
+# If "" or "*" operator will watch every namespace
+watchedNamespace: "*"
```
SeriyBg, line 24: Since the user can specify the coma-separated list of values, it may be better to have it as a list of values. IMO, it would be more user-friendly

SeriyBg (approved):

helm-charts/hazelcast-platform-operator/values.yaml
mtyazici: I also thought about converting it into a list, my main concern was to give a value like this
```
watchedNamespaces: [ "*" ]
```
But if this is OK with you, I can change it in the last PR about this epic.

SeriyBg: This one seems fine to me 😃

cagric0 (approved):

PR #590, opened 2023-01-25
Decrease the image build time and add option to re-run tests with rep…

cheels:
## Description

- Add the option to rerun the failed test with the ability to re-generate an allure report. Retention policy for the report: 5 days. After this period becomes impossible to re-run the tests.
- Decreased time for image build from 3.40m up to 30s
- Changed ttl.sh registry to Google Artifact Registry with default removing policy up to 5 days. The cleanup workflow is here https://github.com/hazelcast/DevOps/pull/245

### DIFF
--- a/.github/workflows/backup-e2e-gke.yaml
+++ b/.github/workflows/backup-e2e-gke.yaml
@@ -7,6 +7,10 @@ on:
 env:
   GCP_PROJECT_ID: hazelcast-33
   GKE_ZONE: europe-west1-b
+  NEW_RELIC_API_KEY: ${{ secrets.NEW_RELIC_API_KEY }}
+  GAR_REGION: us-east1
+  GAR_PROJECT: hazelcast-33
+  GAR_REPO: hazelcast-platform-operator
 
 jobs:
   create-gke-cluster:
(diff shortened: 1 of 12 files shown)


.github/workflows/backup-e2e-gke.yaml
```diff
@@ -51,24 +55,62 @@ jobs:
...
-      IMG: ${{ steps.build-img.outputs.IMG }}
+      IMG: ${{ steps.get-image-tag.outputs.IMG }}
     steps:
-      - name: Checkout
-        uses: actions/checkout@v3
+      - uses: actions/checkout@v3
+
+      - name: Set Up Docker Buildx
+        id: buildx
+        uses: docker/setup-buildx-action@master
```
mtyazici, line 64: Should we use master?

.github/workflows/backup-e2e-gke.yaml
```diff
@@ -51,24 +55,62 @@ jobs:
...
+        with:
+          install: true
 
-      - name: Build and push image to ttl.sh
-        id: build-img
+      - name: Cache Docker Layers
+        uses: actions/cache@v3.2.3
+        with:
+          path: /tmp/.buildx-cache
+          key: ${{ runner.os }}-multi-buildx-${{ hashFiles('Dockerfile','main.go','api/**','controllers/**','internal/**','licenses/**','**/go.mod','**/go.sum') }}
```
mtyazici, line 72: It is removed from cache after some time right?

.github/workflows/backup-e2e-gke.yaml
```diff
@@ -109,6 +151,13 @@ jobs:
           project_id: ${{ env.GCP_PROJECT_ID }}
           location: ${{ env.GKE_ZONE }}
 
+      - name: Authenticate to GAR
+        uses: docker/login-action@v2
+        with:
+          registry: us-east1-docker.pkg.dev
+          username: _json_key
```
mtyazici, line 158: Do we need this step to download the image?

.github/workflows/generate-test-report.yaml
```diff
@@ -126,10 +126,3 @@ jobs:
         run: |
           source .github/scripts/utils.sh
           cleanup_page_publish_runs ${GITHUB_REPOSITORY} "pages-build-deployment" ${{ github.run_number }} $CLEANUP_TIMEOUT
-
-      - name: Delete Test Artifact
-        uses: geekyeggo/delete-artifact@v2
-        with:
```
mtyazici, line 132: Do we not not need artifacts to be deleted?

mtyazici (approved):
I have a few questions other than that LGTM :+1:

.github/workflows/backup-e2e-gke.yaml
cheels: We can use the exact version.

.github/workflows/backup-e2e-gke.yaml
cheels: No, we don't need to log in because only pushing requires authentication. Pulling is public access.

.github/workflows/generate-test-report.yaml
cheels: From now, we don't need it because it will remove all artifacts and then it will be impossible to re-generate the allure report. 
So I added a retention policy and the artifact would remove automatically.

cagric0 (approved):

PR #591, opened 2023-01-26
[CN-713] Add multinamespace support for the operator

mtyazici:
## Description

Operator is now able to watch multiple namespaces. Made necessary configuration changes in Helm charts to be able to still use namespace scoped roles when multiple namespaced operator is enabled. 

## User Impact

Operator is able to watch multiple namespaces. Helm configuration is available in `watchedNamespaces  field which takes comma separated namespaces as string.


helm-charts/hazelcast-platform-operator/templates/role.yaml
```diff
@@ -1,6 +1,7 @@
-{{- $watchOwnNamespace := (eq .Release.Namespace .Values.watchedNamespace) -}}
-{{- $watchAllNamespace := (has .Values.watchedNamespace (list "" "*" )) -}}
+{{- $watchOwnNamespace := (eq .Release.Namespace .Values.watchedNamespaces) -}}
```
cagric0, line 1: We don't need to separate `ownNamespace` from other namespaces that the operator will watch depending on the watchedNamespaces field (which can include operator's own namespace). We could always create a role for `operatorNamespaceRules`. And in for loop create roles for all the namespaces given with the `watchedNamespaces` field.

Also, I would use two different files for `operatorNamespaceRules`(role.yaml) and `watchedNamespaceRules`(namespacerole.yaml).

mtyazici: I made the changes :+1:

cagric0 (approved):
LGTM! Good enhancement!

PR #596, opened 2023-01-27
Added retry option for grafana steps in case of installation error

cheels:
## Description

If some Grafana steps failed, only after 3 attempts it will fail.


cagric0 (approved):

cheels:
For now, there is a decision to close the PR until the issue appears again.

PR #601, opened 2023-01-30
Add functions to deploy/cleanup operator using OLM

cheels:
## Description

- Added 2 makefile commands to deploy and cleanup operator using OLM bundle

The run command looks like this:
> `make olm-deploy NS=ns-olm MODE=OwnNamespace`  where NS - is a random namespace name and MODE - is one of the available installation modes AllNamespaces|OwnNamespace| SingleNamespace="ns-olm"
It will create a bundle using an image uploaded to ttl.sh service

To cleanup the deployment you should specify the same namespace.
> `make cleanup-olm NS=ns-olm`

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -313,6 +313,23 @@ bundle: operator-sdk manifests kustomize yq ## Generate bundle manifests and met
 	sed -i  "s|createdAt: REPLACE_DATE|createdAt: \"$$(date +%F)T11:59:59Z\"|" bundle/manifests/hazelcast-platform-operator.clusterserviceversion.yaml
 	$(OPERATOR_SDK) bundle validate ./bundle --select-optional suite=operatorframework
 
+olm-deploy: operator-sdk ## Deploying Operator with OLM bundle
+	@$(eval CONTAINER_IMAGE=ttl.sh/$(shell uuidgen | tr "[:upper:]" "[:lower:]"):4h)
+	@$(eval BUNDLE_IMAGE=ttl.sh/$(shell uuidgen | tr "[:upper:]" "[:lower:]"):4h)
+	@$(eval VERSION=1.0.0)
+	$(MAKE) docker-build-ci IMG=$(CONTAINER_IMAGE) VERSION=$(VERSION)
+	$(MAKE) docker-push IMG=$(CONTAINER_IMAGE)
+	$(MAKE) bundle IMG=${CONTAINER_IMAGE} VERSION=$(VERSION)
+	@printf "  com.redhat.openshift.versions: v4.8\n  operators.operatorframework.io.bundle.channel.default.v1: alpha" >> ./bundle/metadata/annotations.yaml
+	docker build -f bundle.Dockerfile -t ${BUNDLE_IMAGE} .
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -313,6 +313,23 @@ bundle: operator-sdk manifests kustomize yq ## Generate bundle manifests and met
...
+	@$(eval BUNDLE_IMAGE=ttl.sh/$(shell uuidgen | tr "[:upper:]" "[:lower:]"):4h)
+	@$(eval VERSION=1.0.0)
+	$(MAKE) docker-build-ci IMG=$(CONTAINER_IMAGE) VERSION=$(VERSION)
+	$(MAKE) docker-push IMG=$(CONTAINER_IMAGE)
+	$(MAKE) bundle IMG=${CONTAINER_IMAGE} VERSION=$(VERSION)
+	@printf "  com.redhat.openshift.versions: v4.8\n  operators.operatorframework.io.bundle.channel.default.v1: alpha" >> ./bundle/metadata/annotations.yaml
+	docker build -f bundle.Dockerfile -t ${BUNDLE_IMAGE} .
+	docker push ${BUNDLE_IMAGE}
+	$(KUBECTL) create namespace $(NS)
+	operator-sdk run bundle ${BUNDLE_IMAGE} --namespace=$(NS) --timeout=10m --install-mode=OwnNamespace
```
mtyazici, line 327: Can we give other install modes? If we do does the operator-sdk takes an input for giving the `targetNamespaces`?

cheels: We can give other modes but as far as I can understand, we don't support different modes because CSV file contains the next:

```
installModes:
  - supported: true
    type: OwnNamespace
  - supported: false
    type: SingleNamespace
  - supported: false
    type: MultiNamespace
  - supported: false
    type: AllNamespaces
```

And here is what the documentation saying:

> install-mode: specifies which supported [installMode](https://github.com/operator-framework/operator-lifecycle-manager/blob/master/doc/design/building-your-csv.md#operator-metadata) should be used to create an OperatorGroup by configuring its spec.targetNamespaces field. The InstallModeType string passed must be marked as “supported” in the CSV being installed.
This option understands the following strings (assuming your CSV does as well):
AllNamespaces: the Operator will watch all namespaces (cluster-scoped Operators). This is the default.
OwnNamespace: the Operator will watch its own namespace (from namespace or the kubeconfig default).
SingleNamespace="my-ns": the Operator will watch a namespace, not necessarily its own.

mtyazici: We have support for all of them, it is in this PR https://github.com/hazelcast/hazelcast-platform-operator/pull/599

cheels: @mtyazici Ok, so we can parametrise it and run it as next command

> make olm-deploy NS=ns-olm MODE=(AllNamespace|OwnNamespace|SingleNamespace="my-ns")
except for **MultiNamespace** because I don't see that MultiNamespace is in the list of allowed modes.

mtyazici: sounds great, thanks :)

mtyazici (approved):

PR #602, opened 2023-01-30
Add pr label checker

cheels:
## Description

Added PR labels checker workflow. When PR is opened and labeled or unlabeled, the workflow will check the PR if one of the required labels is present. If not- the bot will leave a comment and request the change. 

image

### DIFF
--- a/.github/workflows/pr-label-checker.yml
+++ b/.github/workflows/pr-label-checker.yml
@@ -1,0 +1,28 @@
+name: Pull Request Label Checker
+
+on:
+  pull_request_target:
+    types: [opened, labeled, unlabeled, synchronize]
+    branches:
+      - main
+      - 'release/**'
+  pull_request:
+    types: [opened, labeled, unlabeled, synchronize]
+    branches:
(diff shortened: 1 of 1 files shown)


.github/workflows/pr-label-checker.yml
```diff
@@ -0,0 +1,28 @@
...
+
+jobs:
+  label-checker:
+    name: Run Label Checker
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout
+        uses: actions/checkout@v3
+
+      - name: Verify Pull Request Labels
```
mtyazici, line 23: When a PR is created from fork this workflow will run two times, I think we should put an if condition to allow forks to only run `pull_request_target` and PRs from this repo to run only `pull_request`. You can find an example in the PR builder.

cheels: Thanks, I updated!

mtyazici (approved):

hasancelik (approved):

PR #606, opened 2023-02-01
Fix watchedNamespaces variable and manager options

mtyazici:
## Description

`wachedNamespaces` variable in Makefile had missing `s`. Also manager parameter in `setManagerWathedNamespaces` was not a pointer, so changes in the options struct had no effect.

### DIFF
--- a/api/v1alpha1/hazelcast_validation.go
+++ b/api/v1alpha1/hazelcast_validation.go
@@ -92,7 +92,7 @@ func validateLicense(h *Hazelcast) error {
 		err := kubeclient.Get(context.Background(), secretName, &secret)
 		if kerrors.IsNotFound(err) {
 			// we care only about not found error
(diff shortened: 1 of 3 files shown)


api/v1alpha1/hazelcast_validation.go
```diff
@@ -92,7 +92,7 @@ func validateLicense(h *Hazelcast) error {
 		err := kubeclient.Get(context.Background(), secretName, &secret)
 		if kerrors.IsNotFound(err) {
 			// we care only about not found error
-			return errors.New("secret not found")
+			return errors.New("Hazelcast Enterprise licenseKeySecret secret is not found")
```
mtyazici, line 95:
```suggestion
			return errors.New("Hazelcast Enterprise licenseKeySecret is not found")
```

cagric0 (approved):

hasancelik (approved):

PR #607, opened 2023-02-02
[CN-719] Remove hostPath support for persistence

SeriyBg:
## User Impact

Remove the support of `hostPath` for Hazelcast persistence


mtyazici (commented):
Could you also remove parts in the e2e tests that delete some folders in the `/tmp` directory. It should be in `test/e2e/e2e_suite_test.go`.

mtyazici (commented):
I think we should merge the agent side PR first and update the agent version here as well.

SeriyBg:
> I think we should merge the agent side PR first and update the agent version here as well.

Yes, sure. The agent PR is no the review

mtyazici (approved):

hasancelik (approved):

PR #615, opened 2023-02-06
 Improve map deletion from Wan status 

mtyazici:
## Description

Improved map deletion from Wan status when publisher ID was empty. Also improved logs for easier debugging.

### DIFF
--- a/controllers/hazelcast/wanreplication_controller.go
+++ b/controllers/hazelcast/wanreplication_controller.go
@@ -84,7 +84,7 @@ func (r *WanReplicationReconciler) Reconcile(ctx context.Context, req ctrl.Reque
 
 	if !wan.GetDeletionTimestamp().IsZero() {
 		if controllerutil.ContainsFinalizer(wan, n.Finalizer) {
-			if err := r.deleteTerminatingMapsFromStatus(ctx, wan); err != nil {
+			if err := r.deleteLeftoverMapsFromStatus(ctx, wan); err != nil {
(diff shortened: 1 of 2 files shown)


controllers/hazelcast/wanreplication_controller.go
```diff
@@ -199,12 +199,21 @@ func (r *WanReplicationReconciler) deleteTerminatingMapsFromStatus(ctx context.C
...
-	for _, m := range hzMaps {
+	for mapWanKey := range wan.Status.WanReplicationMapsStatus {
+		m, ok := publishedMaps[mapWanKey]
+		if !ok {
+			// Delete map without publisherID from status
+			if err := deleteWanMapStatus(ctx, r.Client, wan, mapWanKey); err != nil {
+				return err
+			}
+			continue
+		}
```
cagric0, line 215: Since you remove the map with empty `publisherId` from the status in the [stopWanReplicationMap](https://github.com/hazelcast/hazelcast-platform-operator/pull/615/files#diff-2c75a0b94a1af139aee8a5d8426617b85ee057ea0189636f1fcb0be7a575cbafR246-R248), I think we don't need to change here and an additional check for `publisherId`

cagric0 (commented):
I just have one suggestion. Other than that, LGTM! After you address or give an answer, I'll approve.

cagric0 (approved):

SeriyBg (approved):

PR #630, opened 2023-02-14
Update maploader function

cheels:
## Description

Update map loader function


semihbkgr (approved):
Do not forget to update the names of the e2e test cases.

semihbkgr:
I set the hotbackup slow test "should interrupt external backup process when the hotbackup is deleted" to pending. I am fixing it in another PR.

PR #631, opened 2023-02-15
[CN-718] Add route support for MC

mtyazici:
## User Impact

Users can create route resources in OpenShift environments for MC


controllers/managementcenter/managementcenter_resources.go
```diff
@@ -124,6 +125,51 @@ func (r *ManagementCenterReconciler) reconcileIngress(ctx context.Context, mc *h
 	return err
 }
 
+func (r *ManagementCenterReconciler) reconcileRoute(ctx context.Context, mc *hazelcastv1alpha1.ManagementCenter, logger logr.Logger) error {
+	if platform.GetType() != platform.OpenShift {
+		return nil
```
SeriyBg, line 130: Would you be able to make this scenario more verbose for the user?
Maybe ideally could be to generate the K8s event, but maybe at least some log for the time being would be good

SeriyBg (approved):

controllers/managementcenter/managementcenter_resources.go
mtyazici: Which scenario, if you mean about Route only being enabled for OpenShift we have validation for it.

PR #632, opened 2023-02-15
Update helm-chart-sync.yaml workflow

cheels:


cagric0 (approved):

.github/workflows/helm-chart-sync.yaml
```diff
@@ -22,6 +22,7 @@ jobs:
           git config user.email "devopshelm@hazelcast.com"
           git config user.name "devOpsHelm"
           git subtree add --prefix=charts https://github.com/hazelcast/charts.git master --squash
+          git rm -r charts/stable/hazelcast-platform-operator/*
```
mtyazici, line 25: I did not understand why we put this line, could you explain?

cheels: @mtyazici When you execute `git subtree add` it adds the `hazelcast/charts.git` repo. Then if you remove something from `hazelcast-platform-operator/charts` during development after some time (e.g.helm-charts/hazelcast-platform-operator/templates/role-watched-ns.yaml) then it won't sync with subtree, and all leftover files will be kept in the submodule. It happened last time https://github.com/hazelcast/charts/tree/master/stable/hazelcast-platform-operator/templates

mtyazici: Got it, thanks 👍🏼

mtyazici (approved):

PR #639, opened 2023-02-20
Fixing the stuck e2e tests

semihbkgr:
## Description

It is fixing the stuck e2e tests where the logs are checked. 
It provides the non-blocking function to be called in gomega `Eventually` function, which avoids the test case to be stuck for more than one hour.


cagric0 (approved):

test/e2e/hazelcast_user_code_test.go
```diff
@@ -102,17 +101,16 @@ var _ = Describe("Hazelcast User Code Deployment", Label("custom_class"), func()
...
 			Expect(line).To(ContainSubstring(k + "=" + v))
 		}
-		test.EventuallyInLogs(scanner, 10*Second, logInterval).Should(ContainSubstring(fmt.Sprintf("SimpleStore - Map name is %s", m.MapName())))
-		test.EventuallyInLogs(scanner, 10*Second, logInterval).Should(ContainSubstring("SimpleStore - loading all keys"))
-		test.EventuallyInLogs(scanner, 10*Second, logInterval).Should(ContainSubstring(fmt.Sprintf("SimpleStore - storing key: %d", entryCount-1)))
-
+		test.EventuallyInLogs(logReader, 10*Second, logInterval).Should(ContainSubstring(fmt.Sprintf("SimpleStore - Map name is %s", m.MapName())))
+		test.EventuallyInLogs(logReader, 10*Second, logInterval).Should(ContainSubstring("SimpleStore - loading all keys"))
+		test.EventuallyInLogs(logReader, 10*Second, logInterval).Should(ContainSubstring(fmt.Sprintf("SimpleStore - storing key: %d", entryCount-1)))
+		Expect(logReader.Close()).NotTo(HaveOccurred())
```
SeriyBg, line 113: Maybe it is better to use `defer` to close the `logReader` to avoid leaks when the test fails

SeriyBg (approved):

test/e2e/hazelcast_user_code_test.go
semihbkgr: Yes, it would be better 👍

PR #643, opened 2023-02-22
Add external adress for ingress and route

mtyazici:
## User Impact

Ingress and Route addresses will be present in Management Center external adresses status field.

### DIFF
--- a/controllers/managementcenter/managementcenter_controller.go
+++ b/controllers/managementcenter/managementcenter_controller.go
@@ -6,10 +6,12 @@ import (
 	"time"
 
 	"github.com/go-logr/logr"
+	routev1 "github.com/openshift/api/route/v1"
 	appsv1 "k8s.io/api/apps/v1"
 	corev1 "k8s.io/api/core/v1"
 	"k8s.io/apimachinery/pkg/api/errors"
 	"k8s.io/apimachinery/pkg/runtime"
+	"k8s.io/apimachinery/pkg/types"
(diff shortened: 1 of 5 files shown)


controllers/managementcenter/managementcenter_controller.go
```diff
@@ -158,3 +161,20 @@ func (r *ManagementCenterReconciler) updateLastSuccessfulConfiguration(ctx conte
...
+		externalAddresses = append(externalAddresses, mc.Spec.ExternalConnectivity.Ingress.Hostname+":80")
+	}
+
+	if mc.Spec.ExternalConnectivity.Route.IsEnabled() {
+		route := &routev1.Route{}
+		err := cli.Get(ctx, types.NamespacedName{Name: mc.Name, Namespace: mc.Namespace}, route)
+		if err != nil {
+			return externalAddresses
+		}
+		externalAddresses = append(externalAddresses, route.Spec.Host+":80")
```
kutluhanmetin, line 176: Is it possible to use `route.Spec.Port.String()` instead of `":80"`?

kutluhanmetin (approved):

controllers/managementcenter/managementcenter_controller.go
mtyazici: That port is not external port, it is the port of the service which might be different.

cagric0 (approved):
LGTM!

PR #644, opened 2023-02-22
[CN-746]: Version Diff Tool

kutluhanmetin:
## Description

Prints api spec differences between version numbers.

Example 1 - if you provide versions:
`version-diff-tool -base=5.2 -revision=5.3` will print diff to `diff_open-api-bundle-5.2.yaml_open-api-bundle-5.3.yaml.txt`

Example 2 - if you don't provide versions, it will compare the latest 2 versions:
`version-diff-tool` will print diff to `diff_open-api-bundle-5.5.yaml_open-api-bundle-5.6.yaml.txt`

### DIFF
--- a/tools/version-diff-tool/main.go
+++ b/tools/version-diff-tool/main.go
@@ -1,21 +1,47 @@
 package main
 
 import (
+	"flag"
 	"fmt"
+	"github.com/hashicorp/go-version"
 	"github.com/tidwall/sjson"
 	"io"
 	"net/http"
 	"os"
 	"os/exec"
+	"path/filepath"
 	"sigs.k8s.io/yaml"
 	"strings"
 )
 
 const (
+	// File names
+	outputFileName          = `diff_%s_%s.txt`
+	descriptionlessFileName = `r%s`
+	bundleFileName          = `bundle-%s.yaml`
+	openAPIFileName         = `open-api-%s`
+
+	// Splitters
+	schemaSplitter = `---`
+
+	// Links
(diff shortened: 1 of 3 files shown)


tools/version-diff-tool/main.go
```diff
@@ -0,0 +1,297 @@
...
+	openAPIFileName         = `open-api-%s`
+
+	// Splitters
+	schemaSplitter = `---`
+
+	// Links
+	bundleDownloadLink = `https://repository.hazelcast.com/operator/` + bundleFileName
+
+	// Helm related
+	helmInstall      = `helm install operator hazelcast/hazelcast-platform-operator --dry-run --set installCRDs=true > ` + bundleFileName
```
cagric0, line 31:
```suggestion
	helmInstall      = `helm template operator hazelcast/hazelcast-platform-operator-crds --version=%s > ` + bundleFileName
```

I think we don't need to call the server for rendering so we could use the `template` command instead of `install`.
Also, one potential problem is not setting the chart version; when we have new releases, our helm commands for the operator chart will use the latest version by default. So we need to pass the version explicitly.

cagric0 (changes requested):
I tested the tool and it's working, thanks for the clean output. I have just one change request after that I'll approve. 

One thing about the PR template, if the pr is not related to the release or there is no user impact you can directly remove `user impact` part. Also if the pr is not a release item you should label the PR as `non-release`.

mtyazici (approved):

PR #647, opened 2023-02-28
[CN-675]: JVM GC Configuration

kutluhanmetin:
## Description

Enables to set JVM GC Configuration using Hazelcast Platform Operator.

## User Impact

- Users can enable GC logging.
- Users can set GC (Serial, Parallel, G1)
- Users can pass arbitrary JVM GC Arguments.


api/v1alpha1/hazelcast_types.go
```diff
@@ -611,6 +622,59 @@ func (c *JVMMemoryConfiguration) GetMaxRAMPercentage() string {
...
 
+// JVMGCConfiguration is for configuring JVM Garbage Collector
+type JVMGCConfiguration struct {
+	// Logging enables logging when set to true
+	// +optional
+	Logging *bool `json:"logging,omitempty"`
+
+	// Collector is the Garbage Collector type
+	// +optional
+	Collector *string `json:"collector,omitempty"`
```
mtyazici, line 633: I would suggest we put GCType here and not a string

mtyazici (approved):
LGTM, are we planning to add phone home or integration tests for GC config?

api/v1alpha1/hazelcast_types.go
cagric0: I agree with Muhammet. We should change its type to enum.

cagric0:
> LGTM, are we planning to add phone home or integration tests for GC config?

We should discuss it with the team but I think integration tests are a must-have. I don't have a strong opinion about phone-home.

cagric0 (approved):

PR #653, opened 2023-03-09
[CN-755] [CN-676]: JVM Args Changes

kutluhanmetin:
## Description

* add arbitrary args to jvm
* add validations for arbitrary args to webhook
* add phone home for jvm config
* add integration test for validations

## User Impact

User will be able to set arbitrary JVM arguments


api/v1alpha1/hazelcast_validation.go
```diff
@@ -56,6 +70,50 @@ func validateClusterSize(h *Hazelcast) error {
...
+			}
+		}
+	}
+	return nil
+}
+
+func validateArg(args []string, arg string) error {
+	for _, s := range args {
+		if strings.Contains(s, arg) {
+			return fmt.Errorf(`argument %s is configured twice, please check "Configuring JVM Parameters" document`, arg)
```
hasancelik, line 111: Maybe we can directly mention the precedence here without redirecting the user to the doc?

hasancelik (approved):

PR #661, opened 2023-03-16
[CN-638]: Add near cache support for Map CR

kutluhanmetin:
## Description

Near Cache support for Map CR is added. However, there are some problems with codec and I opened issues for them on core side (continued with default values):
- https://github.com/hazelcast/hazelcast/issues/23978
- https://github.com/hazelcast/hazelcast/issues/23979

## User Impact

User will be able to configure Near Cache for Map CR.
Example:

```
kind: Map
metadata:
  name: mostly-read-map
spec:
  hazelcastResourceName: hazelcast
  nearCache:
      inMemoryFormat: OBJECT
      invalidateOnChange: true
      timeToLiveSeconds: 300
      maxIdleSeconds: 300
      eviction:
        evictionPolicy: NONE
        maxSizePolicy: ENTRY_COUNT
        size: 5000
      cacheLocalEntries: true
```


test/integration/hazelcast_controller_test.go
```diff
@@ -945,6 +945,58 @@ var _ = Describe("Hazelcast controller", func() {
...
+				Expect(ms.NearCache.Name).To(Equal(m.Spec.NearCache.Name))
+				Expect(ms.NearCache.CacheLocalEntries).To(Equal(m.Spec.NearCache.CacheLocalEntries))
+				Expect(ms.NearCache.InvalidateOnChange).To(Equal(m.Spec.NearCache.InvalidateOnChange))
+				Expect(ms.NearCache.MaxIdleSeconds).To(Equal(m.Spec.NearCache.MaxIdleSeconds))
+				Expect(ms.NearCache.InMemoryFormat).To(Equal(m.Spec.NearCache.InMemoryFormat))
+				Expect(ms.NearCache.TimeToLiveSeconds).To(Equal(m.Spec.NearCache.TimeToLiveSeconds))
+				Expect(ms.NearCache.NearCacheEviction.EvictionPolicy).To(Equal(m.Spec.NearCache.NearCacheEviction.EvictionPolicy))
+				Expect(ms.NearCache.NearCacheEviction.Size).To(Equal(m.Spec.NearCache.NearCacheEviction.Size))
+				Expect(ms.NearCache.NearCacheEviction.MaxSizePolicy).To(Equal(m.Spec.NearCache.NearCacheEviction.MaxSizePolicy))
+				Delete(m)
```
SeriyBg, line 997: I guess all the cleanup should happen in `After` functions to prevent interference with other tests

SeriyBg (changes requested):
One small comment

test/integration/hazelcast_controller_test.go
kutluhanmetin: Would you give an example from tests? I see similar uses of `Delete(m)` in almost every controller test.

SeriyBg: You are right, but this seems wrong. The correct approach would be to cleanup in the `AfterSuite` as in the E2E tests. However, I think it can be fixed in a separate PR. Could you please create a ticker for this?

SeriyBg (approved):

test/integration/hazelcast_controller_test.go
kutluhanmetin: okay 👍

PR #670, opened 2023-03-20
Added workflow to test Operator on multiple version

cheels:
## Description

- Added workflow to test the latest Operator on multiple latest HZ and MC versions: 5.1.x and 5.0.x.
- As a result, there is one report will be created: 
image


mtyazici (approved):

.github/workflows/multi-version-tests.yaml
```diff
@@ -0,0 +1,296 @@
+name: Multi Version Tests
+
+on:
+  workflow_dispatch:
+  schedule:
+    - cron: "0 9 */3 * *"
```
hasancelik, line 6: Running this build every three days does not make sense. Because we know that this suite will fail with the current latest versions until they release them, we can keep the suite manual trigger based and run it before the releases to see if additional failures exist.

hasancelik (changes requested):

hasancelik (approved):

PR #671, opened 2023-03-22
[CN-770] Cleaning up leftover volume from Grafana agents

semihbkgr:

### DIFF
--- a/.github/workflows/cleanup-namespace.yaml
+++ b/.github/workflows/cleanup-namespace.yaml
@@ -9,16 +9,7 @@ on:
         type: string
         required: true
       namespaces:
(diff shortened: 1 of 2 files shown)


.github/workflows/cleanup-namespace.yaml
```diff
@@ -0,0 +1,95 @@
+name: Install Grafana Agent
```
cheels, line 1: Is it correct name?

cheels (changes requested):

.github/workflows/cleanup-namespace.yaml
semihbkgr: It is not correct.

mtyazici (approved):

PR #683, opened 2023-03-30
[CN-795]  Improve validation logic for multiple errors 

mtyazici:
## User Impact

Webhooks and Reconciler validation now return multiple errors instead of the first found


api/v1alpha1/hazelcast_validation.go
```diff
@@ -32,118 +34,73 @@ var BlackListProperties = map[string]struct{}{
...
 	}
 
 	if err := validateLicense(h); err != nil {
-		return err
+		allErrs = append(allErrs, err)
 	}
 
 	if err := validateTLS(h); err != nil {
-		return err
+		return append(allErrs, err)
```
semihbkgr, line 58: Why returning errors after TLS validation?

api/v1alpha1/cache_validation.go
```diff
@@ -1,39 +1,17 @@
...
-	}
-	return nil
+func ValidateCacheSpecCurrent(c *Cache, h *Hazelcast) error {
+	var allErrs field.ErrorList
+	allErrs = appendIfNotNil(allErrs, ValidateAppliedPersistence(c.Spec.PersistenceEnabled, h))
+	allErrs = appendIfNotNil(allErrs, ValidateAppliedNativeMemory(c.Spec.InMemoryFormat, h))
+	if len(allErrs) == 0 {
+		return nil
+	}
+	return kerrors.NewInvalid(schema.GroupKind{Group: "hazelcast.com", Kind: "Cache"}, c.Name, allErrs)
```
semihbkgr, line 16: As a suggestion: It would be better to have common parametrized function as we use this logic in multiple places.

api/v1alpha1/data_structure.go
semihbkgr, line 1: `Map` doesn't implement `DataStructure` interface after these changes.

mtyazici: Map did not implement the DataStructure interface before as well. We are not using the interface for the map anywhere so we can implement it when we decide to use it.

api/v1alpha1/hazelcast_validation.go
mtyazici: It is wrong, you are right. I will fix it 👍🏼

api/v1alpha1/cache_validation.go
mtyazici: We use this logic in two ways. We always check for length and then sometimes we create a new error and other times we return the exact error. If you have a suggestion for both of them, we can change it.

semihbkgr (approved):

api/v1alpha1/common_validation.go
```diff
@@ -0,0 +1,102 @@
...
+		allErrs = append(allErrs, field.Forbidden(field.NewPath("spec"), "cannot be updated"))
+	}
+	if len(allErrs) == 0 {
+		return nil
+	}
+	return kerrors.NewInvalid(schema.GroupKind{Group: "hazelcast.com", Kind: GetKind(obj)}, obj.GetName(), allErrs)
+}
+
+func isDSSpecUnchanged(obj client.Object) (bool, error) {
+	lastSpec, ok := obj.GetAnnotations()[n.LastSuccessfulSpecAnnotation]
```
SeriyBg, line 79: Same question here. Maybe by using the `old` object we can omit covering multiple negative scenarios

api/v1alpha1/common_validation.go
```diff
@@ -0,0 +1,102 @@
...
+		return nil
+	}
+
+	s, ok := h.ObjectMeta.Annotations[n.LastSuccessfulSpecAnnotation]
+	if !ok {
+		return field.InternalError(field.NewPath("spec"), fmt.Errorf("hazelcast resource %s is not successfully started yet", h.Name))
+	}
+
+	lastSpec := &HazelcastSpec{}
+	err := json.Unmarshal([]byte(s), lastSpec)
```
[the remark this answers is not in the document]
mtyazici: You are right. But using the old object we cannot know if the old object was successfully applied. So we might be validating on things actually have not taken place. I agree this scenario is an unlikely one since the user has to be fast enough. If you don't see much value in it. I can change the logic.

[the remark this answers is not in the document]
mtyazici: I am not sure if I understood what you meant. Could you elaborate that?

My point in using the lastSuccessfulSpecAnnotation is like follows: When a successfully applied CR is updated. But updated CR is somehow fails at the reconcilation logic. Then what spec should we use to validate when the user updates the failed spec again. Should we validate against the successfully applied one, which is in the annotation, or the failed one, which in this case is the old object. I believe using the annotation brings more value.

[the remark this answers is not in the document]
SeriyBg: Oh, I see. You are right!

SeriyBg (approved):

PR #687, opened 2023-04-03
[CN-470] [CN-809] Add possibility to submit the Jet job using JetJob CR

SeriyBg:
## User Impact

Users can submit the JetJob CR from an existing Jar file that will run the Jet job


controllers/hazelcast/jetjob_controller.go
```diff
@@ -0,0 +1,376 @@
...
+			"name", jj.Name, "namespace", jj.Namespace, "state", jj.Status.Phase)
+		return
+	}
+
+	if jj.Status.Phase.IsFinished() {
+		r.removeJobFromChecker(jj)
+		logger.Info("JetJob is already finished.",
+			"name", jj.Name, "namespace", jj.Namespace, "state", jj.Status.Phase)
+		return
+	}
```
mtyazici, line 92: Can we also put Create and Update validation in the reconciler function? I know there will be too much validation functions called in the reconciler. Maybe we can reduce the number like we do in Hazelcast and other CRs.

controllers/hazelcast/jetjob_status.go
```diff
@@ -0,0 +1,103 @@
...
+	sc.m.Lock()
+	if sc.checkInProgress {
+		sc.m.Unlock()
+		return
+	}
+	sc.checkInProgress = true
+	sc.ticker = time.NewTicker(10 * time.Second)
+	go func(f jetJobUpdateFunc) {
+		for ; true; <-sc.ticker.C { //trick to make first tick happen immediately
+			logger.V(util.DebugLevel).Info("Checking JetJob statuses by ticker...")
```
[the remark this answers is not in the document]
SeriyBg: Yes, and we are calling it on line 67 (`sc.ticker.Stop()`) when there are no active jobs anymore

controllers/hazelcast/jetjob_controller.go
SeriyBg: It is there, [line 111 ](https://github.com/hazelcast/hazelcast-platform-operator/pull/687/files#diff-1aace02e07945ad4a1241da1b94b4fce9a5193ff0dd15a8005a70fb0dd9ba50eR111) as an example of the update validation. Or am I missing the point?

mtyazici (approved):

PR #688, opened 2023-04-03
Increased timeout for multiple BackupTests

cheels:
## Description

This PR fixes two tests for Backup

### DIFF
--- a/test/e2e/helpers_test.go
+++ b/test/e2e/helpers_test.go
@@ -7,6 +7,7 @@ import (
 	"encoding/xml"
 	"fmt"
 	"io"
+	"k8s.io/apimachinery/pkg/watch"
 	"log"
 	"net"
 	"net/http"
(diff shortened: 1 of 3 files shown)


kutluhanmetin (approved):

test/e2e/helpers_test.go
```diff
@@ -183,6 +183,7 @@ func DeletePod(podName string, gracePeriod int64, lk types.NamespacedName) {
 		if err != nil {
 			log.Fatal(err)
 		}
+		Sleep(1 * Minute)
```
SeriyBg, line 186: TBH, I'm uncomfortable when there is `Sleep` in the code. It will increase the run time by 1 minute even if the pods are deleted faster. Maybe we can implement some sort of [pod watcher](https://stackoverflow.com/a/71202736/5604676)?

SeriyBg (approved):

PR #691, opened 2023-04-05
[CN-813]: Fix Advanced Network Default WAN Configuration

kutluhanmetin:
## Description

- Added default scenario tests to advanced networking.
- If user does not provide any advanced networking config, WAN Replication will be available through Hazelcast service on port 5710.
- If user configures wan replication but does not provide a serviceType, than serviceType is set to `LoadBalancer`. 

## User Impact
- Users will be able to list client external addresses and wan addresses as separate columns when they type `kubectl get hazelcast`. Example output:

```
NAME              STATUS    MEMBERS   EXTERNAL-ADDRESSES    WAN-ADDRESSES
hazelcast-44      Running   3/3                             34.67.64.241:5720,34.67.64.241:5721,34.67.64.241:5722,34.67.64.241:5723,34.67.64.241:5724,104.154.82.70:5750,104.154.82.70:5751
hazelcast-55      Running   3/3
hazelcast-first   Running   3/3       35.193.112.105:5701   35.193.112.105:5710
```
- As shown in the `hazelcast-44` example, when multiple wan configurations applied, then all of the endpoints are list under `WAN-ADDRESSES`.
- If users creates a Hazelcast without expose externally enabled, they will not see any external and wan addresses as shown with `hazelcast-55`.
- If users create a Hazelcast with expose externally, but don't provide any WAN configurations, then default WAN configuration will be applied as shown with `hazelcast-first`.

- If user does not provide any advanced networking config, WAN Replication will be available through Hazelcast service on port 5710.

- If user configures wan replication but does not provide a serviceType, than serviceType is set to `LoadBalancer`.


controllers/hazelcast/hazelcast_resources.go
```diff
@@ -343,14 +338,20 @@ func (r *HazelcastReconciler) reconcileService(ctx context.Context, h *hazelcast
...
 	}
 
-	opResult, err := util.CreateOrUpdate(ctx, r.Client, service, func() error {
+	opResult, err := util.CreateOrUpdateForce(ctx, r.Client, service, func() error {
 		service.Spec.Type = serviceType(h)
 		service.Spec.Ports = util.EnrichServiceNodePorts(hazelcastPort(), service.Spec.Ports)
+
+		// append default wan port to HZ Discovery Service if use did not configure
+		if len(h.Spec.AdvancedNetwork.WAN) == 0 {
+			service.Spec.Ports = append(service.Spec.Ports, defaultWANPort())
```
mtyazici, line 352: Could you put WANPort logic into `hazelcastPort` function. With this set up there will be infinite trigger between operator and service controllers since operator will override the NodePort as 0.

hasancelik (approved):

mtyazici (approved):

PR #710, opened 2023-04-14
remove TLS enterprise validation in webhook

semihbkgr:
## Description

removing TLS enterprise validation in Webhook


mtyazici (approved):
I thought TLS was an enterprise feature. Is that wrong?

semihbkgr:
Yes, It is an enterprise feature but we don't have this kind of checks for other enterprise features in Webhook.
Additionally, this check is not correct. I have created a task about adding LicenseKey check in Webhook for each enterprise features.
You can take a look at CN-827

PR #713, opened 2023-04-14
[CN-817] Add support for OpenSSL TLS engine

semihbkgr:
## Description

adding OpenSSL TLS engine option

## User Impact

OpenSSL TLS engine support

```yaml
kind: Hazelcast
metadata:
  name: hazelcast
spec:
  ...
  tls:
    secretName: tls-secret
    type: OpenSSL
```

values of `tls.type`:
    - BasicSSL (default)
    - OpenSSL


api/v1alpha1/hazelcast_types.go
```diff
@@ -789,8 +789,34 @@ type ClientServerSocketEndpointConfig struct {
...
 	SecretName string `json:"secretName,omitempty"`
+
+	// Type of TLS configuration
+	// +kubebuilder:default:="BasicSSL"
+	// +optional
+	Type TLSType `json:"type,omitempty"`
+}
+
+func (t *TLS) IsEnabled() bool {
+	return t != nil && t.SecretName != ""
```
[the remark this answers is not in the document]
semihbkgr: We can have 3 options in that case:
-  consider as TLS disabled
- fail on validation
- create a TLS secret automatically

I think it is better to fail on validation.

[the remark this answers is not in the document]
SeriyBg: My vote would also be going to the validation. As, IMO, other two might be confusing 😃

controllers/hazelcast/hazelcast_resources.go
```diff
@@ -858,44 +862,69 @@ func hazelcastBasicConfig(h *hazelcastv1alpha1.Hazelcast) config.Hazelcast {
...
-			jksPath  = path.Join(n.HazelcastMountPath, "hazelcast.jks")
+			jksPath  = path.Join(n.HazelcastMountPath, n.TLSBasicSSLJKSFile)
 			password = "hazelcast"
 		)
-		// require MTLS for member-member communication
-		cfg.AdvancedNetwork.MemberServerSocketEndpointConfig.SSL = config.SSL{
+		// require mTLS for member-member communication
+		member = config.SSL{
 			Enabled:          pointer.Bool(true),
 			FactoryClassName: "com.hazelcast.nio.ssl.BasicSSLContextFactory",
```
cagric0, line 885: It's better to create constants and use them.Also you could you these constants in the tests.

test/e2e/hazelcast_test.go
```diff
@@ -5,15 +5,13 @@ import (
...
 	"k8s.io/apimachinery/pkg/types"
 
+	hazelcastcomv1alpha1 "github.com/hazelcast/hazelcast-platform-operator/api/v1alpha1"
+	hazelcastconfig "github.com/hazelcast/hazelcast-platform-operator/test/e2e/config/hazelcast"
 	. "github.com/onsi/ginkgo/v2"
 	. "github.com/onsi/gomega"
 	"k8s.io/apimachinery/pkg/api/errors"
-
-	hazelcastcomv1alpha1 "github.com/hazelcast/hazelcast-platform-operator/api/v1alpha1"
-	hazelcastconfig "github.com/hazelcast/hazelcast-platform-operator/test/e2e/config/hazelcast"
```
cagric0, line 16: Could you reorder the imports? We should be consistent about them. 

- Built-in libs

- Other than built-in and k8s libs

- K8s libs 

- Local imports

controllers/managementcenter/managementcenter_resources.go
```diff
@@ -325,11 +325,16 @@ func (r *ManagementCenterReconciler) reconcileSecret(ctx context.Context, mc *ha
 	opResult, err := util.CreateOrUpdateForce(ctx, r.Client, secret, func() error {
 		files := make(map[string][]byte)
 		for _, cluster := range mc.Spec.HazelcastClusters {
-			keystore, err := hazelcastKeystore(ctx, r.Client, mc, cluster.TLS.SecretName)
-			if err != nil {
-				return err
+
+			// Management Center cluster TLS configuration
+			// TLS.Type field is just ignored. We continue with 'BasicSSL'
+			if cluster.TLS.IsEnabled() {
```
cagric0, line 331: From now on, we are not depending on the secretName field for checking if TLS enabled or not. So you've added validation for TLS on the Hazelcast CR. However, it could lead us to a problem for the mancenter CR where we pass TLS config for each cluster. So checking just TLS `IsEnabled` not enough. In this code snippet, when we don't pass `secretName`, hazelcastKeyStore is returning a broken keystore instead of an error. So we should add validation here and also to the MC webhook if it's suitable.

controllers/hazelcast/hazelcast_resources.go
```diff
@@ -602,40 +601,45 @@ func (r *HazelcastReconciler) isServicePerPodReady(ctx context.Context, h *hazel
...
+				if h.Spec.TLS.IsBasicSSLEnabled() {
+					keystore, err := hazelcastKeystore(ctx, r.Client, h)
+					if err != nil {
+						return err
+					}
+					configSecret.Data[n.TLSBasicSSLJKSFile] = keystore
 				}
-				cm.Data["hazelcast.jks"] = keystore
+				// If the TLS type is `OpenSSL`, the cert and key values from given secret is not copied into the secret.
+				// The given TLS secret is directly mounted into the ManagementCenter container.
```
cagric0, line 632: This comment is confusing for me, why are we mounting it into the MC container?

controllers/hazelcast/hazelcast_resources.go
```diff
@@ -858,44 +862,69 @@ func hazelcastBasicConfig(h *hazelcastv1alpha1.Hazelcast) config.Hazelcast {
...
+			FactoryClassName: "com.hazelcast.nio.ssl.OpenSSLEngineFactory",
+			Properties:       NewOpenSSLProperties(crtPath, keyPath, "TLS", tls.MutualAuthentication),
 		}
 	}
-	return cfg
+	return
 }
 
-func NewSSLProperties(path, password, protocol string, auth v1alpha1.MutualAuthentication) config.SSLProperties {
+func NewBasicSSLProperties(path, password, protocol string, auth hazelcastv1alpha1.MutualAuthentication) config.SSLProperties {
```
SeriyBg, line 913: It seems like for this `func`, the protocol is always `TLS`. Maybe let's remove this argument?

api/v1alpha1/hazelcast_validation.go
```diff
@@ -128,20 +128,33 @@ func validateLicense(h *Hazelcast) *field.Error {
 }
 
 func validateTLS(h *Hazelcast) *field.Error {
-	// make sure secret exists
-	if h.Spec.TLS.SecretName != "" {
+
+	if h.Spec.TLS.IsEnabled() {
```
SeriyBg, line 132: I would suggest following the [align the happy path to the left](https://medium.com/@matryer/line-of-sight-in-code-186dd7cdea88) rule here. Like this:
```go
if !h.Spec.TLS.IsEnabled() {
  return nil
}
```

controllers/hazelcast/hazelcast_resources.go
semihbkgr: That `func` is from other PR. I only changes the name. I think it is not a big deal.

semihbkgr:
https://hazelcast.slack.com/archives/C01JU7ZJYGP/p1683701597796119

PR #723, opened 2023-04-26
[CN-810] Add possibility to specify the JAR file to be downloaded in the JetJob CR

SeriyBg:
## User Impact

Users can specify the bucket from the JAR file that will be downloaded in the JetJob CR


test/integration/jetjob_webhook_test.go
```diff
@@ -54,14 +58,43 @@ var _ = Describe("Hazelcast webhook", func() {
...
+			Expect(k8sClient.Create(context.Background(), jj)).Should(Succeed())
+
+			jj.Spec.BucketConfiguration = &hazelcastv1alpha1.BucketConfiguration{
+				Secret:    "my-secret",
+				BucketURI: "gs://my-bucket",
+			}
+			err := k8sClient.Update(context.Background(), jj)
+			Expect(err).Should(And(
+				MatchError(ContainSubstring("spec.bucketConfiguration")),
+				MatchError(ContainSubstring("Forbidden: field cannot be added or removed")),
```
hasancelik, line 96: What if this job is canceled/failed or finished?

config/samples/_v1alpha1_jetjob_dynamic.yaml
```diff
@@ -0,0 +1,19 @@
+apiVersion: hazelcast.com/v1alpha1
+kind: JetJob
+metadata:
+  labels:
+    app.kubernetes.io/name: jetjob
```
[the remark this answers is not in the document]
SeriyBg: We don't. It's autogenerated by the newest version of the `operator-sdk`

test/integration/jetjob_webhook_test.go
hasancelik: > If the job is canceled, we should still not allow users to change the configs, as this might mean that a new JAR needs to be downloaded and run. This is basically a new Job.

👍 

> Good point about the failed Job; however, we cannot know if it failed because the Jar wasn't downloaded or the job failed during the execution unless we introduce a new status, i.e. FailedDowloadingFile

I applied JetJob CR but my bucket is not reachable with my creds so I need to fix them and then create a new CR with a new name. At least, if we can keep this status as an internal status then take action based on that. WDYT?

SeriyBg: Good catch. I've introduced a new status.Now there is separate `FailedDuringExection` and `Failed` states. For the `Failed` state it is possible to update the JetJob CR

hasancelik (approved):

PR #727, opened 2023-05-04
Kubeclient setup in Integration test

semihbkgr:
## Description

Setting up `kubeclient` in Integration test as well.

### DIFF
--- a/main.go
+++ b/main.go
@@ -98,7 +98,7 @@ func main() {
 	}
 
 	if err := mgr.Add(kubeclient.Setup(mgr.GetClient())); err != nil {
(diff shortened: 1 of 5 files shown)


semihbkgr:
In the validation functions, we are using 'kubeclient' package to get resources.
eg. `err := kubeclient.Get(context.Background(), secretName, &secret)`

As we are not setting up `kubeclient` in Integration test, 'kubeclient' uses the default noop k8s (controller runtime) client.
This causes us to not test validations in Integration tests.

SeriyBg (approved):

PR #740, opened 2023-05-16
[CN-812] Rename secret to secretName for ManagementCenter and HotBackup

SeriyBg:
## User Impact

The field licenseKeySecret in ManagementCenterSpec was renamed to licenseKeySecretName.
The field secret in HotBackupSpec was renamed to secretName.

### DIFF
--- a/test/e2e/config/managementcenter/config.go
+++ b/test/e2e/config/managementcenter/config.go
@@ -127,7 +127,7 @@ var (
 			Spec: hazelcastv1alpha1.ManagementCenterSpec{
 				Repository:                 *mcRepo,
 				Version:                    "not-exists",
-				DeprecatedLicenseKeySecret: licenseKey(ee),
+				LicenseKeySecretName: licenseKey(ee),
 			},
 		}
 	}
--- a/test/e2e/config/managementcenter/config.go
+++ b/test/e2e/config/managementcenter/config.go
@@ -125,8 +125,8 @@ var (
 				Labels:    lbls,
 			},
 			Spec: hazelcastv1alpha1.ManagementCenterSpec{
-				Repository:                 *mcRepo,
-				Version:                    "not-exists",
+				Repository:           *mcRepo,
+				Version:              "not-exists",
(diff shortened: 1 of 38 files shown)


test/e2e/config/managementcenter/config.go
```diff
@@ -125,9 +125,9 @@ var (
 				Labels:    lbls,
 			},
 			Spec: hazelcastv1alpha1.ManagementCenterSpec{
-				Repository:       *mcRepo,
-				Version:          "not-exists",
-				LicenseKeySecret: licenseKey(ee),
+				Repository:                 *mcRepo,
+				Version:                    "not-exists",
+				DeprecatedLicenseKeySecret: licenseKey(ee),
```
semihbkgr, line 130:
```suggestion
				LicenseKeySecretName: licenseKey(ee),
```

test/e2e/helpers_test.go
```diff
@@ -905,7 +905,7 @@ func restoreConfig(hotBackup *hazelcastcomv1alpha1.HotBackup, useBucketConfig bo
 		return hazelcastcomv1alpha1.RestoreConfiguration{
 			BucketConfiguration: &hazelcastcomv1alpha1.BucketConfiguration{
 				BucketURI:  hotBackup.Status.GetBucketURI(),
-				SecretName: hotBackup.Spec.Secret,
+				SecretName: hotBackup.Spec.SecretName,
```
semihbkgr, line 908:
```suggestion
				SecretName: hotBackup.Spec.GetSecretName(),
```

semihbkgr (commented):
Maybe we can consider to set values to new fields taking values from their corresponding deprecated fields in reconciler loop.

SeriyBg:
> Maybe we can consider to set values to new fields taking values from their corresponding deprecated fields in reconciler loop.

This is a good option. But for this, I think we also need to have the `defaulter` webhook. So I will create a JIRA ticket to implement it in a separate task.

PR #744, opened 2023-05-24
Added HZ, MC versions as input params to workflow and skip jet job tests for 5.1.X version 

cheels:
## Description

Since we do not support JetJobs for 5.1.X it should be disabled in the MultiVersion workflow.

### DIFF
--- a/test/e2e/hazelcast_jetjob_test.go
+++ b/test/e2e/hazelcast_jetjob_test.go
@@ -4,6 +4,7 @@ import (
 	"context"
 	"fmt"
 	"os"
+	"strings"
 	. "time"
 
 	. "github.com/onsi/ginkgo/v2"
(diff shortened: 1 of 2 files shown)


test/e2e/hazelcast_jetjob_test.go
```diff
@@ -25,6 +27,10 @@ var _ = Describe("Hazelcast JetJob", Label("JetJob"), func() {
 		if !useExistingCluster() {
 			Skip("End to end tests require k8s cluster. Set USE_EXISTING_CLUSTER=true")
 		}
+		version := os.Getenv("HZ_VERSION")
+		if strings.Contains(version, "5.1.6") || strings.Contains(version, "5.0.5") {
```
SeriyBg, line 31: Maybe it's better to use `semver` package for comparing the versions? Otherwise we need to change this one for every minor version release

PR #750, opened 2023-06-05
[CN-864] Add support for custom configs using ConfigMap

SeriyBg:
## User Impact

Added possibility to add custom Hazelcast configuration specified in ConfigMap via `customConfigCmName` field in Hazelcast CR


cagric0 (approved):
LGTM!

controllers/hazelcast/hazelcast_resources.go
```diff
@@ -697,9 +697,41 @@ func hazelcastConfig(ctx context.Context, c client.Client, h *hazelcastv1alpha1.
...
+		}
+		hzWrapper := make(map[string]interface{})
+		hzWrapper["hazelcast"] = cfgMap
+		return yaml.Marshal(hzWrapper)
+	}
+
 	return yaml.Marshal(config.HazelcastWrapper{Hazelcast: cfg})
 }
 
+func mergeCustomConfig(cfg *config.Hazelcast, cstCfg map[string]interface{}) error {
```
semihbkgr, line 720: This function basically merges the `config` into the `custom config`. Would be good to rename the function and change the parameters order.

test/integration/hazelcast_config_test.go
semihbkgr, line 1: We did you create a new test file instead of residing them in `hazelcast_test.go`

semihbkgr (changes requested):
What would happen if user gives incorrect Hazelcast config fields in custom config map. Should we check custom config fields in validation?

controllers/hazelcast/hazelcast_resources.go
SeriyBg: Would you happen to have any suggestions for the name?
Why change the order of the parameters?

test/integration/hazelcast_config_test.go
SeriyBg: This test file is focused only on the Hazelcast config. I'm not a big fan of having huge files with everything in it, and think that it's better to separate them for a better readbility

SeriyBg:
> What would happen if user gives incorrect Hazelcast config fields in custom config map. Should we check custom config fields in validation?

The decision on not to validate the custom config. Here you can find the reasoning: https://hazelcast.atlassian.net/wiki/spaces/CN/pages/4513398825/Custom+config+support+for+Hazelcast+CR#Should-we-validate-the-custom-config%3F

controllers/hazelcast/hazelcast_resources.go
semihbkgr: my suggestion: `func mergeConfig(cstCfg map[string]interface{},cfg *config.Hazelcast) error`

semihbkgr (approved):

semihbkgr (approved):

PR #751, opened 2023-06-05
[CN-801] Implement JetJobSnapshot CR

semihbkgr:
## Description

Implementation of JetJobSnapshot CR

## User Impact

Add JetJobSnapshot CR that provides a way to export snapshots of JetJob CRs.


controllers/hazelcast/jetjobsnapshot_controller.go
```diff
@@ -0,0 +1,267 @@
...
+					withJetJobSnapshotState(hazelcastv1alpha1.JetJobSnapshotFailed))
+				return
+			}
+			logger.Info("Snapshot is exported successfully", "name", jjs.SnapshotName())
+			setCreationTime(exportSnapshotCtx, r.Client, n, t)              //nolint:errcheck
+			updateJetJobSnapshotStatusRetry(exportSnapshotCtx, r.Client, n, //nolint:errcheck
+				withJetJobSnapshotState(hazelcastv1alpha1.JetJobSnapshotExported))
+		}(k)
+	} else {
+		logger.Info("Exporting process has already been started", "name", jjs.SnapshotName(), "cancelJob", jjs.Spec.CancelJob)
```
SeriyBg, line 224: Probably for such log it's better set the debug level

controllers/hazelcast/jetjobsnapshot_controller.go
```diff
@@ -0,0 +1,267 @@
...
+		}
+		jjs.ObjectMeta.Annotations[n.LastSuccessfulSpecAnnotation] = string(jjss)
+
+		return r.Client.Update(ctx, jjs)
+	})
+}
+
+func (r *JetJobSnapshotReconciler) executeFinalizer(ctx context.Context, jjs *hazelcastv1alpha1.JetJobSnapshot, logger logr.Logger) error {
+	k := types.NamespacedName{Name: jjs.Name, Namespace: jjs.Namespace}
+	r.exportingSnapshotMap.Delete(k)
```
SeriyBg, line 251: Is it right to delete the entry before the finalizer is deleted?

SeriyBg (changes requested):

cagric0 (commented):
Overall LGTM. Thanks for the huge effort you put for making the operator great with the jet job!!! I left some minor comments, after you address them and Sergii's comments. I'll approve it.

controllers/hazelcast/jetjobsnapshot_controller.go
semihbkgr: Nice catch 👍. I think it isn't right, but we are doing it in multiple CRs finalizer such as Hazelcast CR, JetJob CR. 
We can create a new ticket referring the order of the operations in the finalizers. What do you think?

api/v1alpha1/jetjobsnapshot_validation.go
```diff
@@ -0,0 +1,75 @@
...
+		return nil
+	}
+	var parsed JetJobSnapshotSpec
+	if err := json.Unmarshal([]byte(last), &parsed); err != nil {
+		return []*field.Error{field.InternalError(field.NewPath("spec"), fmt.Errorf("error parsing last JetJobSnapshot spec for update errors: %w", err))}
+	}
+	return ValidateJetJobSnapshotNonUpdatableFields(jjs.Spec, parsed)
+}
+
+func ValidateJetJobSnapshot(h *Hazelcast) error {
```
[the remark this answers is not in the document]
semihbkgr: Yes, but since the scope of the functions is package level in Go, so I cannot set so generic name to this function.
I have only conformed the naming convention we have in JetJob validation.

```go
func ValidateJetConfiguration(h *Hazelcast) error {
```

We can find more generic solution or better naming convention for these functions.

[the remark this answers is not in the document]
semihbkgr: `ValidateHazelcasEnterprise` would be better 👍

semihbkgr:
why the new commit are not included in the PR

semihbkgr:
migrated to #791 as the new incoming commits are not loaded in the PR.

controllers/hazelcast/jetjobsnapshot_controller.go
semihbkgr: We missed the case where getting an error after deleting finalizer. Just imagine we removed the finalizer first, then get en error while applying actual finalizer operations. Then we cannot stand a chance to run reconciler and finalizer again.

SeriyBg: But what if during the finalizer operations an error occurs, but the `exportingSnapshotMap` entry is already deleted?

semihbkgr: If that happens, then the reconciler will be triggered again and the finalizer is called again. What is the problem?

PR #765, opened 2023-06-19
Include additional fields of JetJob CR to wide output

semihbkgr:
## Description

kubectl output with `-o wide` flag displays `FAILURETEXT` and `SUSPENSIONCAUSE` fields

### DIFF
--- a/api/v1alpha1/jetjob_types.go
+++ b/api/v1alpha1/jetjob_types.go
@@ -116,6 +116,8 @@ func (jjs JetJobStatusPhase) IsSuspended() bool {
 // +kubebuilder:printcolumn:name="Id",type="string",JSONPath=".status.id",description="ID of the JetJob"
 // +kubebuilder:printcolumn:name="SubmissionTime",type="string",JSONPath=".status.submissionTime",description="Time when the JetJob was submitted"
 // +kubebuilder:printcolumn:name="CompletionTime",type="string",JSONPath=".status.completionTime",description="Time when the JetJob was completed"
+// +kubebuilder:printcolumn:name="FailureText",type="string",priority=1,JSONPath=".status.failureText",description="The text about failure of the JetJob"
+// +kubebuilder:printcolumn:name="SuspensionCause",type="string",priority=1,JSONPath=".status.suspensionCause",description="The reason why the JetJob is suspended"
 // +kubebuilder:resource:shortName=jj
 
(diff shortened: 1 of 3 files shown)


api/v1alpha1/jetjob_types.go
```diff
@@ -116,6 +116,8 @@ func (jjs JetJobStatusPhase) IsSuspended() bool {
 // +kubebuilder:printcolumn:name="Id",type="string",JSONPath=".status.id",description="ID of the JetJob"
 // +kubebuilder:printcolumn:name="SubmissionTime",type="string",JSONPath=".status.submissionTime",description="Time when the JetJob was submitted"
 // +kubebuilder:printcolumn:name="CompletionTime",type="string",JSONPath=".status.completionTime",description="Time when the JetJob was completed"
+// +kubebuilder:printcolumn:name="FailureText",type="string",priority=1,JSONPath=".status.failureText",description="The text about failure of the JetJob"
+// +kubebuilder:printcolumn:name="SuspensionCause",type="string",priority=1,JSONPath=".status.suspensionCause",description="The reason why the JetJob is suspended"
```
SeriyBg, line 120: I thought the idea was to have only one field for the status, where `FailureText` or `SuspensionCause` will be printed. I found it weird to have `FailureText` as a permanent output, as if we're always expecting a failure

SeriyBg (changes requested):

api/v1alpha1/jetjob_types.go
semihbkgr: What do you mean by 'permanent output' ?

SeriyBg: Whenever one runs `kubectl get jetjob -owide` there will be these fields, even if there is no failure message. And at least one of these fields will always be empty. Instead, it's better to have one `message` field for both fields

semihbkgr:
Due the variety of the message format (Failure Text might refer to validation error or the error from Hazelcast cluster), It cannot be displayed in a good shape on the console.

api/v1alpha1/jetjob_types.go
cagric0: I agree with Sergii. It should be consistent with other CRs.

cagric0 (changes requested):
Other than Sergii's comment, there is nothing to add. After you update, I'll directly approve the PR.

PR #770, opened 2023-06-22
[CN-891] Fix update procedure

SeriyBg:
- Replaced all the `+optional` structs in the specs to a pointer
- Add defaulter webhooks and mutation function to remove empty structs from saved configs
- Move validation for the TLS secretName to the webhook/reconcileer logic


cheels (approved):

api/v1alpha1/hazelcast_webhook.go
```diff
@@ -38,3 +42,32 @@ func (r *Hazelcast) ValidateDelete() error {
...
+
+func (r *Hazelcast) Default() {
+	if r.Spec.LicenseKeySecretName == "" && r.Spec.DeprecatedLicenseKeySecret != "" {
+		r.Spec.LicenseKeySecretName = r.Spec.DeprecatedLicenseKeySecret
+		r.Spec.DeprecatedLicenseKeySecret = ""
+	}
+	r.defaultOptionalToNil()
+}
+
+func (r *Hazelcast) defaultOptionalToNil() {
```
hasancelik, line 54: Do we need to update this method every time a new embedded struct is defined in Hazelcast CR?

SeriyBg: No, as long as these structs are pointers. We just need not to make the same error :)

hasancelik (approved):

PR #771, opened 2023-06-23
[CN-882] Delete JetJob CR when its Hazelcast CR is deleted

semihbkgr:
## Description

Finalizer of Hazelcast CR deletes its dependent JetJob CRs. 
When the JetJob CR is deleted, if its Hazelcast CR has marked to be deleted, (we assume that the deletion operation of JetJob CR is triggered by Hazelcast CR finalizer), then the JetJob finalizer does't cancel the actual job. In this way, the jet jobs in the cluster can be recovered by `losslesRestart` in the backup of HotRestart.

## User Impact

When Hazelcast CR object is getting deleted, its dependent JetJob CR objects are deleted as well without canceling the actual jobs. In case of restoring the cluster from a backup, the jobs were in the cluster can be recovered by `losslesRestart`.

### DIFF
--- a/controllers/hazelcast/jetjob_controller.go
+++ b/controllers/hazelcast/jetjob_controller.go
@@ -374,16 +374,18 @@ func (r *JetJobReconciler) executeFinalizer(ctx context.Context, jj *hazelcastv1
 		return nil
 	}
 
-	hzNn := types.NamespacedName{Name: jj.Spec.HazelcastResourceName, Namespace: jj.Namespace}
+	hzNn := types.NamespacedName{
+		Name:      jj.Spec.HazelcastResourceName,
+		Namespace: jj.Namespace,
+	}
 	hz := &hazelcastv1alpha1.Hazelcast{}
 	err := r.Client.Get(ctx, hzNn, hz)
 	if err != nil {
 		if apiErrors.IsNotFound(err) {
 			logger.Info("Hazelcast resource not found. Ignoring since object must be deleted")
-			return nil
-		} else {
-			return err
+			return r.removeCheckerAndFinalizer(ctx, jj)
 		}
+		return err
 	}
 
 	// If Hazelcast CR is getting deleted, the finalizer doesn't cancel the actual job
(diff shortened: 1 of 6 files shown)


controllers/hazelcast/jetjob_controller.go
```diff
@@ -384,21 +403,14 @@ func (r *JetJobReconciler) executeFinalizer(ctx context.Context, jj *hazelcastv1
 	return nil
 }
 
-func (r *JetJobReconciler) stopJetExecution(ctx context.Context, jj *hazelcastv1alpha1.JetJob, logger logr.Logger) error {
+func (r *JetJobReconciler) stopJetExecution(ctx context.Context, jj *hazelcastv1alpha1.JetJob, h *hazelcastv1alpha1.Hazelcast, logger logr.Logger) error {
```
SeriyBg, line 406: It seems that we only need Hazelcast's `types.NamespacedName`, not the entire `hazelcastv1alpha1.Hazelcast` object here

semihbkgr: That is true and also would be good to have an utility function that returns `types.NamespacedName` of the given resource. We are doing this same thing lots of time in the source code.

controllers/hazelcast/jetjob_controller.go
```diff
@@ -373,9 +373,28 @@ func (r *JetJobReconciler) executeFinalizer(ctx context.Context, jj *hazelcastv1
...
-	if err := r.stopJetExecution(ctx, jj, logger); err != nil {
-		return fmt.Errorf("failed to remove finalizer: %w", err)
+
+	hzNn := types.NamespacedName{Name: jj.Spec.HazelcastResourceName, Namespace: jj.Namespace}
+	hz := &hazelcastv1alpha1.Hazelcast{}
+	err := r.Client.Get(ctx, hzNn, hz)
+	if err != nil {
+		if apiErrors.IsNotFound(err) {
+			logger.Info("Hazelcast resource not found. Ignoring since object must be deleted")
+			return nil
```
SeriyBg, line 383: Don't we need to remove the finalizer and `removeJobFromChecker` in this case?

SeriyBg (changes requested):

controllers/hazelcast/jetjob_controller.go
semihbkgr: It is better removing the finalizer when Hazelcast CR is not found 👍

SeriyBg (approved):

PR #783, opened 2023-07-03
[CN-891] fix mutation detection for CR update

SeriyBg:
Update the CRD when mutated in the reconciler

### DIFF
--- a/controllers/hazelcast/mutate/mutate.go
+++ b/controllers/hazelcast/mutate/mutate.go
@@ -1,6 +1,8 @@
 package mutate
 
 import (
+	"reflect"
+
 	hazelcastv1alpha1 "github.com/hazelcast/hazelcast-platform-operator/api/v1alpha1"
 	n "github.com/hazelcast/hazelcast-platform-operator/internal/naming"
 )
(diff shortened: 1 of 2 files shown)


controllers/hazelcast/mutate/mutate.go
```diff
@@ -10,6 +12,10 @@ func HazelcastSpec(h *hazelcastv1alpha1.Hazelcast) (mutated bool) {
 		h.Spec.Repository = n.HazelcastEERepo
 		mutated = true
 	}
+	oldHz := &h
 	h.Default()
+	if !reflect.DeepEqual(oldHz, &h) {
```
semihbkgr, line 17: return statement can be simplified

semihbkgr (changes requested):
It is just small suggestion.

controllers/hazelcast/mutate/mutate.go
SeriyBg: `mutated` variable might be changed above. IMO, such simplification will make the code less readable

semihbkgr: I meant something like this
```go
oldHz := &h
h.Default()
if h.Spec.GetLicenseKeySecretName() != "" && h.Spec.Repository == n.HazelcastRepo {
		h.Spec.Repository = n.HazelcastEERepo
		return true
}
return !reflect.DeepEqual(oldHz, &h)
```
not big deal, It is still good anyway 👍

semihbkgr (approved):
LGTM

hasancelik (approved):

PR #784, opened 2023-07-03
[CN-871] Added integration tests for hz clusters' tls configuration on mc

cagric0:
## Description

Added integration test for the tls configuration validation on MC. Validation is added in [another PR](https://github.com/hazelcast/hazelcast-platform-operator/pull/755).

### DIFF
--- a/test/integration/managementcenter_test.go
+++ b/test/integration/managementcenter_test.go
@@ -476,7 +476,8 @@ var _ = Describe("ManagementCenter CR", func() {
 					},
 				}}
 
-				Expect(k8sClient.Create(context.Background(), mc)).Should(HaveOccurred())
+				Expect(k8sClient.Create(context.Background(), mc)).
+					Should(MatchError(ContainSubstring("Management Center Cluster config TLS Secret name is empty")))
 			})
 
 			It("should error when secretName does not exist", Label("fast"), func() {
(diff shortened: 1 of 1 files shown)


test/integration/managementcenter_test.go
```diff
@@ -462,6 +462,38 @@ var _ = Describe("ManagementCenter CR", func() {
...
+				}
+				mc.Spec.HazelcastClusters = []hazelcastv1alpha1.HazelcastClusterConfig{{
+					Name:    "dev",
+					Address: "dummy",
+					TLS: &hazelcastv1alpha1.TLS{
+						SecretName: "",
+					},
+				}}
+
+				Expect(k8sClient.Create(context.Background(), mc)).Should(HaveOccurred())
```
SeriyBg, line 479: Maybe it would be better to assert an error message as well? Just to make sure that the error is the one that we expect

SeriyBg (approved):
LGTM, with one small comment

test/integration/managementcenter_test.go
semihbkgr, line 479: Checking the error message might be good. It guarantees that the error is caused by the desired reason.
Example:
```go
Expect(k8sClient.Create(context.Background(), mc)).
					Should(MatchError(ContainSubstring("<error message>")))
```

test/integration/managementcenter_test.go
```diff
@@ -462,6 +462,38 @@ var _ = Describe("ManagementCenter CR", func() {
...
+				}
+				mc.Spec.HazelcastClusters = []hazelcastv1alpha1.HazelcastClusterConfig{{
+					Name:    "dev",
+					Address: "dummy",
+					TLS: &hazelcastv1alpha1.TLS{
+						SecretName: "notfound",
+					},
+				}}
+
+				Expect(k8sClient.Create(context.Background(), mc)).Should(HaveOccurred())
```
semihbkgr, line 495: same as above

semihbkgr (approved):
LGTM with small suggestion 👍

PR #785, opened 2023-07-03
Added migration test

cheels:
## Description

Added migration test as part of release process. It will test migration from previous stable release to latest snapshot. 
In case of 10 message of "level":"error" in a row then it will return the logs from operator and fail the job.

### DIFF
--- a/.github/workflows/test-and-release.yaml
+++ b/.github/workflows/test-and-release.yaml
@@ -15,8 +15,86 @@ env:
   GAR_REGION: us-east1
   GAR_PROJECT: hazelcast-33
   GAR_REPO: hazelcast-platform-operator
+  GKE_ZONE: us-east1-b
 
 jobs:
+  migration-test:
+    name: Run migration test
+    runs-on: ubuntu-latest
+    env:
+      GCP_NETWORK: operator-test-network
+    steps:
+      - name: Authenticate to GCP
+        uses: 'google-github-actions/auth@v1.1.1'
+        with:
+          credentials_json: ${{ secrets.GKE_SA_KEY }}
+
+      - name: Create GKE cluster
+        run: |-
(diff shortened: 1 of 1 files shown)


.github/workflows/test-and-release.yaml
```diff
@@ -15,8 +15,110 @@ env:
...
+          metadata:
+            name: hazelcast
+          spec:
+            scheduling:
+              nodeSelector:
+                topology.kubernetes.io/region: ${{ env.GAR_REGION }}
+            nativeMemory:
+              allocatorType: POOLED
+              size: 100M
+            clusterSize: 1
```
cagric0, line 77: I don't know if it matters but I think it's nice to make it more realistic with clusterSize 3

cagric0 (approved):
LGTM! I just left one minor suggestion.

.github/workflows/test-and-release.yaml
```diff
@@ -15,8 +15,110 @@ env:
...
+                size: 5000
+          EOF
+          
+          echo "Checking Hazelcast CR before upgrade"
+          kubectl wait --for=jsonpath='{.status.phase}'=Running --timeout=120s hazelcast/hazelcast
+          echo "Checking Map CR before upgrade"
+          kubectl wait --for=jsonpath='{.status.state}'=Success --timeout=120s map/persistent
+          
+          helm upgrade operator hazelcast/hazelcast-platform-operator --version=$LATEST_VERSION --set=installCRDs=true --set=phoneHomeEnabled=false --wait
+          echo "Checking Hazelcast CR after upgrade"
```
hasancelik, line 112: Could you add a check for the operator pod as well?

hasancelik (changes requested):

.github/workflows/test-and-release.yaml
cheels: Done ✅

hasancelik (approved):

PR #791, opened 2023-07-05
[CN-801] JetJobSnapshot CR Implementation

semihbkgr:
## Description

Implementation of JetJobSnapshot CR

This PR is copy of #751

## User Impact

Add JetJobSnapshot CR that provides a way to export snapshots of JetJob CRs.


api/v1alpha1/cache_types.go
```diff
@@ -32,6 +31,10 @@ type CacheSpec struct {
 	// +kubebuilder:default:=BINARY
 	// +optional
 	InMemoryFormat InMemoryFormatType `json:"inMemoryFormat,omitempty"`
+
+	// EventJournal specifies event journal configuration of the Cache
+	// +optional
+	EventJournal *EventJournal `json:"eventJournal,omitempty"`
```
SeriyBg, line 37: Is this change related to the JetJobSnapshot? Because if it is not, this change will be missing from the release notes

api/v1alpha1/jetjobsnapshot_validation.go
```diff
@@ -0,0 +1,69 @@
...
+	kerrors "k8s.io/apimachinery/pkg/api/errors"
+	"k8s.io/apimachinery/pkg/runtime/schema"
+	"k8s.io/apimachinery/pkg/util/validation/field"
+
+	n "github.com/hazelcast/hazelcast-platform-operator/internal/naming"
+)
+
+func ValidateJetJobSnapshotSpecCreate(jjs *JetJobSnapshot) error {
+	var allErrs field.ErrorList
+	if jjs.Spec.JetJobResourceName == "" {
```
SeriyBg, line 17: The field's `//+required` tag is enough for the validation

controllers/hazelcast/jetjob_controller.go
```diff
@@ -198,12 +198,27 @@ func (r *JetJobReconciler) applyJetJob(ctx context.Context, job *hazelcastv1alph
...
+		jjsnn := types.NamespacedName{Name: job.Spec.InitialSnapshotResourceName, Namespace: job.Namespace}
+		jjs := hazelcastv1alpha1.JetJobSnapshot{}
+		err := r.Client.Get(ctx, jjsnn, &jjs)
+		if err != nil {
+			logger.Error(err, "Error on getting JetJobSnapshot custom resource",
+				"snapshotResourceName", job.Spec.InitialSnapshotResourceName)
+			return r.updateStatus(ctx, jjnn, failedJetJobStatus(err))
+		}
+		if jjs.Status.CreationTime.IsZero() {
+			err := fmt.Errorf("JetJobSnapshot '%s' has not exported", job.Spec.InitialSnapshotResourceName)
```
SeriyBg, line 211:
```suggestion
			err := fmt.Errorf("JetJobSnapshot '%s' is not exported", job.Spec.InitialSnapshotResourceName)
```

controllers/hazelcast/jetjobsnapshot_controller.go
```diff
@@ -0,0 +1,273 @@
...
+			withJetJobSnapshotFailedState(err.Error()))
+	}
+
+	if err := hazelcastv1alpha1.ValidateHazelcastLicenseKey(hz); err != nil {
+		logger.Info("License Key is not set", "name", hzNn)
+		return updateJetJobSnapshotStatus(ctx, r.Client, jjs, recoptions.Error(err),
+			withJetJobSnapshotFailedState(err.Error()))
+	}
+
+	if hz.Status.Phase != hazelcastv1alpha1.Running {
```
SeriyBg, line 154: Isn't this check redundant? We already check for the JetJob state, and it won't be `Running` if the Hazelcast cluster isn't running

SeriyBg (changes requested):

api/v1alpha1/cache_types.go
semihbkgr: Should we create task or add in release notes?

api/v1alpha1/jetjobsnapshot_validation.go
semihbkgr: Are you sure?
Without this check, I could have created snapshot with empty `JetJobResourceName`.

SeriyBg: Then you can use `// +kubebuilder:validation:MinLength:=`, this will work for sure

api/v1alpha1/cache_types.go
SeriyBg: We should probably manually add to release notes. If it's already added, then there is no need to remove it, but we have to include in the release notes

api/v1alpha1/jetjobsnapshot_validation.go
cagric0: I am gonna ask you the same question Semih. Are you sure :) 
It shouldn't be like you said. For example, I tried it for map CR. I tried to create map CR without passing `hazelcastResourceName` and I got the following error.

`error: error validating "map.yaml": error validating data: ValidationError(Map.spec): missing required field "hazelcastResourceName" in com.hazelcast.v1alpha1.Map.spec; if you choose to ignore these errors, turn validation off with --validate=false`

cagric0 (approved):

api/v1alpha1/cache_types.go
semihbkgr: I have informed Alex about that. We will add EventJournal in RN manually after release 👍

PR #799, opened 2023-07-07
[CN-879] fail JetJob if Hazelcast cluster is not running

SeriyBg:
## User Impact

Creating the JetJob for the Hazelcast cluster that is not Running will result in Failed state


semihbkgr (approved):
LGTM 👍. Cannot we test this case in integration tests?

SeriyBg:
> LGTM 👍. Cannot we test this case in integration tests?

Nope, as the JetJob won't run in the integration test

PR #882, opened 2023-09-18
[CN-967]Increased number of parallel jobs

cheels:
## Description

Increased the number of parallel jobs from 5 to 10 and decreased the number of GINKGO_PARALLEL_PROCESSES up to 2 for Kind tests

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -211,7 +211,7 @@ endif
 GINKGO_PARALLEL_PROCESSES ?= 4
 
 test-e2e-split-kind: generate fmt vet ginkgo ## Run end-to-end tests on Kind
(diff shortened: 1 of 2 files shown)


Makefile
```diff
@@ -211,7 +211,7 @@ endif
 GINKGO_PARALLEL_PROCESSES ?= 4
```
hasancelik, line 212: Creating a separate environment for KinD env might be a good idea.

cheels: Done!

hasancelik (approved):
LGTM! Thanks for the change 👍 🙂

PR #901, opened 2023-10-13
[CN-1032] Updated workflow to release Operator with patch version

cheels:
## Description

Updated workflow to release Operator with patch version

### DIFF
--- a/.github/workflows/docs-release.yaml
+++ b/.github/workflows/docs-release.yaml
@@ -3,10 +3,10 @@ on:
   workflow_dispatch:
     inputs:
       RELEASE_VERSION:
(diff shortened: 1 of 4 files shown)


semihbkgr (approved):
LGTM👍 One question: don't we need to set specific operator image version in `values.yaml`?

cagric0 (commented):
We don't need to remove tag even we should not remove it. We just need to update the tag before the release. 
For example:
tag: latest-snapshot -> 5.10.0

cheels:
> LGTM👍 One question: don't we need to set specific operator image version in `values.yaml`?

I already removed this change.

cagric0 (approved):

PR #932, opened 2023-11-07
[CN-1011] Added a soak test for pause/restart functionality

cheels:
## Description
Added soak test for pause/ restart feature
- It will run for 12 hours.
- During this period it should perform 10 pauses and restart
- As for the data - the total PVC size is 14500. 
- Before each pause and restart it should fill the map equally and slowly not at once. For example if 12 hours and 10 restarts, then it will restart each 72 minutes. And during this short period it should fill 1450 mb slowly to emulate the real usage of cluster.
- Restart/Restore the cluster with the 5.24 version’s persistence data.
 - Restart/Restore the cluster with the latest version’s persistence data.


hasancelik (approved):
We need to get Viridian team's review before merge this PR.

test/e2e/helpers_test.go
```diff
@@ -329,58 +329,59 @@ func FillTheMapData(ctx context.Context, lk types.NamespacedName, unisocket bool
...
 			}
-			return hzMap.Size(ctx)
-		}, timeout, 5*Second).Should(Equal(mapSize))
-	})
+			size, err := hzMap.Size(ctx)
+			if err != nil {
+				Fail("ailed to get map size")
+			}
+			if size == mapSize {
+				log.Printf("Success: '%s' map is of size '%d' using lookup name '%s'", mapName, mapSize, lk.Name)
```
semihbkgr, line 367: Shouldn't we need to break the loop in this condition?

semihbkgr: https://github.com/hazelcast/hazelcast-platform-operator/pull/951

Issue #904, opened 2023-10-17
Parallel Pod Management Policy Causing Startup Issues


afoley-st:
Hi there! We have a hazelcast deployment that starts up using the platform operator. On first boot of the cluster (every time), it fails to start up any member of the hazelcast cluster with the following logs:

```shell
########################################
# JAVA=/usr/bin/java
# JAVA_OPTS=--add-modules java.se --add-exports java.base/jdk.internal.ref=ALL-UNNAMED --add-opens java.base/java.lang=ALL-UNNAMED --add-opens java.base/sun.nio.ch=ALL-UNNAMED --add-opens java.management/sun.management=ALL-UNNAMED --add-opens jdk.management/com.sun.management.internal=ALL-UNNAMED -Dhazelcast.logging.type=log4j2 -Dlog4j.configurationFile=file:/opt/hazelcast/config/log4j2.properties -Dhazelcast.config=/opt/hazelcast/config/hazelcast-docker.xml -Djet.custom.lib.dir=/opt/hazelcast/custom-lib -Djava.net.preferIPv4Stack=true -XX:MaxRAMPercentage=80.0 -XX:MaxGCPauseMillis=5 -Dhazelcast.config=/data/hazelcast/hazelcast.yaml -XX:+UseContainerSupport
# CLASSPATH=/opt/hazelcast/*:/opt/hazelcast/userCode/bucket/*:/opt/hazelcast/userCode/urls/*:/opt/hazelcast/lib:/opt/hazelcast/lib/*:/opt/hazelcast/bin/user-lib:/opt/hazelcast/bin/user-lib/*
########################################
{"time":"2023-10-17T12:09:20,330", "logger": "com.hazelcast.internal.config.AbstractConfigLocator", "level": "INFO", "msg": "Loading configuration '/data/hazelcast/hazelcast.yaml' from System property 'hazelcast.config' "}
{"time":"2023-10-17T12:09:20,335", "logger": "com.hazelcast.internal.config.AbstractConfigLocator", "level": "INFO", "msg": "Using configuration file at /data/hazelcast/hazelcast.yaml "}
{"time":"2023-10-17T12:09:23,079", "logger": "com.hazelcast.internal.config.MemberDomConfigProcessor", "level": "WARN", "msg": "<instance> tag is deprecated, use <jet> tag directly for configuration. "}
[...]

PR #939, opened 2023-11-23
DO NOT MERGE - set publishNotReadyAddresses to true

semihbkgr:
We can set `publishNotReadyAddresses` to true in the discovery service. This ensures that all Hazelcast member pods are considered ready by the service. Consequently, Istio can correctly route both inbound and outbound traffic as expected, similar to the behavior observed when the `PodManagementPolicy` is set to `OrderedReady` in a `StatefulSet`.

This change is intended to resolve the issue we are encountering in the Istio setup, specifically issue #904.


semihbkgr:
The issue that is explained in the PR comment has already been fixed by the Istio in its new version and back-ported to two older versions.
https://github.com/hazelcast/hazelcast-platform-operator/issues/904#issuecomment-1824438926

PR #944, opened 2023-12-04
Retry on creating new manager at the start

semihbkgr:
## Description

It will prevent the operator pods from crashing and restarting at startup, especially in an Istio environment when the Envoy sidecar proxy is not yet ready to accept inbound requests.


cagric0 (approved):

semihbkgr:
Restarting has no impact on the operator or the setup. Just let the operator fail and restart.

PR #948, opened 2023-12-13
Fix soak tests 

cheels:
## Description

Fix WaitForMapSize() function


hasancelik (approved):

semihbkgr (commented):
Why we ignore the error propagated from `WaitForMapSize` function?

PR #952, opened 2023-12-22
[CN-1106] Fix EKS workflow adding Ingress Rule to the Security Group of Nodes

semihbkgr:
## Description

Add inbound rule to the SecurityGroup of cluster nodes.

### DIFF
--- a/.github/workflows/e2e-eks.yaml
+++ b/.github/workflows/e2e-eks.yaml
@@ -35,6 +35,7 @@ jobs:
         id: create-cluster
         run: |-
           CLUSTER_NAME="operator-e2e-test-${GITHUB_SHA::8}-${{ github.run_number }}"
+          echo "CLUSTER_NAME=${CLUSTER_NAME}" >> $GITHUB_ENV
           echo "CLUSTER_NAME=${CLUSTER_NAME}" >> $GITHUB_OUTPUT
           eksctl create cluster \
             --name "${CLUSTER_NAME}" \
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-eks.yaml
```diff
@@ -36,21 +36,40 @@ jobs:
...
+          SECURITY_GROUP_ID=$(aws eks describe-cluster --region=${AWS_REGION} --name ${CLUSTER_NAME} --query=cluster.resourcesVpcConfig.clusterSecurityGroupId --output=text)
+
+          echo "SecurityGroupID: ${SECURITY_GROUP_ID}"
+
+          aws ec2 authorize-security-group-ingress \
+            --region=${AWS_REGION} \
+            --group-id=${SECURITY_GROUP_ID} \
+            --cidr=0.0.0.0/0 \
+            --protocol=tcp \
+            --port=30000-32767
```
hasancelik, line 72: As far as I can understand, it is not possible to define port range for NodePort --> https://github.com/aws/containers-roadmap/issues/1361 

I searched for a simple workaround but I could not find a clean one 🙂

hasancelik (approved):

.github/workflows/e2e-eks.yaml
semihbkgr: Yes, and it is not possible using `eksctl` as well.

kutluhanmetin (approved):

PR #958, opened 2023-12-29
[CN-1106] Fix AKS workflow creating NSG rule to allow external connections

semihbkgr:
## Description

Create NetworkSecurityGroup inbound rule to accept external connections to the nodes.

### DIFF
--- a/.github/workflows/e2e-aks.yaml
+++ b/.github/workflows/e2e-aks.yaml
@@ -43,8 +43,28 @@ jobs:
           inlineScript: |
             CLUSTER_NAME="operator-e2e-test-${GITHUB_SHA::8}-${{ github.run_number }}"
             echo "CLUSTER_NAME=${CLUSTER_NAME}" >> $GITHUB_OUTPUT
-            az aks create --resource-group ${AZURE_RESOURCE_GROUP} --name "${CLUSTER_NAME}" \
-              --node-count 3 --generate-ssh-keys
+            az aks create \
+              --resource-group=${AZURE_RESOURCE_GROUP} \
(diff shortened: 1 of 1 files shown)


.github/workflows/e2e-aks.yaml
```diff
@@ -43,8 +43,28 @@ jobs:
...
-            az aks create --resource-group ${AZURE_RESOURCE_GROUP} --name "${CLUSTER_NAME}" \
-              --node-count 3 --generate-ssh-keys
+            az aks create \
+              --resource-group=${AZURE_RESOURCE_GROUP} \
+              --name="${CLUSTER_NAME}" \
+              --node-count=3 \
+              --enable-node-public-ip \
+              --generate-ssh-keys > output.json
+            
+            NRG=$(cat output.json | jq -r .nodeResourceGroup)
```
SeriyBg, line 53: Do we need to create this temporary file? Maybe we can directly use the pipe to `jq`?
```suggestion
            NRG=$(az aks create \
              --resource-group=${AZURE_RESOURCE_GROUP} \
              --name="${CLUSTER_NAME}" \
              --node-count=3 \
              --enable-node-public-ip \
              --generate-ssh-keys | jq -r .nodeResourceGroup)
```

SeriyBg (changes requested):

.github/workflows/e2e-aks.yaml
semihbkgr: Yes, we can simply use pipe to extract `nodeResourceGroup` from output of `az aks create` 👍

SeriyBg (approved):

PR #975, opened 2024-01-16
Check deletion for service-per-pod services

semihbkgr:
## Description

It prevent us to use deleted services which might be leftovers from previous hazelcast cluster.

### DIFF
--- a/controllers/hazelcast/hazelcast_resources.go
+++ b/controllers/hazelcast/hazelcast_resources.go
@@ -689,8 +689,8 @@ func (r *HazelcastReconciler) isServicePerPodReady(ctx context.Context, h *hazel
 	for i := 0; i < int(*h.Spec.ClusterSize); i++ {
 		s := &v1.Service{}
 		err := r.Client.Get(ctx, client.ObjectKey{Name: servicePerPodName(i, h), Namespace: h.Namespace}, s)
-		if err != nil {
-			// Service is not created yet
+		if err != nil || s.GetDeletionTimestamp() != nil {
+			// Service is not created yet or marked as deleted
 			return false
 		}
 		if s.Spec.Type == v1.ServiceTypeLoadBalancer {


controllers/hazelcast/hazelcast_resources.go
```diff
@@ -689,8 +689,8 @@ func (r *HazelcastReconciler) isServicePerPodReady(ctx context.Context, h *hazel
 	for i := 0; i < int(*h.Spec.ClusterSize); i++ {
 		s := &v1.Service{}
 		err := r.Client.Get(ctx, client.ObjectKey{Name: servicePerPodName(i, h), Namespace: h.Namespace}, s)
-		if err != nil {
-			// Service is not created yet
+		if err != nil || s.GetDeletionTimestamp() != nil {
```
hasancelik, line 692: Could you please provide me with details on how you encountered this issue? Is this a recurring problem?

Maybe we need to fix our cleanup logic?

semihbkgr: When the Hazelcast CR is deleted, the operator doesn't block until the service-per-pod services are deleted. These services are deleted by Kubernetes GC in the background, triggered by the removal of their owner (Hazelcast CR in our case). The deletion of services might take some time, especially when the service type is a load balancer (as the cloud provider needs to delete the underlying load balancer resource in its finalizer).

If you delete the Hazelcast CR with externally exposed load balancer and then reapply the same YAML without waiting for the service-per-pod services to be deleted, the operator and Kubernetes discovery plugin will continue to use the old service-per-pod services marked as deleted, which will eventually be removed.

hasancelik: Thanks for the detailed explanation 🙂 

> then reapply the same YAML without waiting for the service-per-pod services to be deleted

It is not a common scenario for users. They will not apply Hazelcast CR multiple times consecutively. Even if they do, we do not need to fix it at the code level; just a document might be enough. WDYT?

I show some resistance to this additional check because this check might affect other core stuff.

cagric0: Are we closing the PR?

PR #988, opened 2024-02-02
Refactor HotBackup persistence validation

semihbkgr:
Use the `HotBackup` object as the source object of the error instead of the `Hazelcast` object.

https://github.com/hazelcast/hazelcast-platform-operator/pull/981#discussion_r1475710007


semihbkgr:
@SeriyBg 

The error message: 
> HotBackup.hazelcast.com "hot-backup" is invalid: spec.persistenceEnabled: Invalid value: false: Persistence must be enabled at Hazelcast

We are using `HotBackup` as source object but using `path` (spec.persistenceEnabled) of `Hazelcast` resource. Is it ok?

SeriyBg (approved):

PR #1024, opened 2024-02-16
Log platform info

semihbkgr:


hasancelik:
@semihbkgr Why do we need this PR? Could you update the description?

PR #1041, opened 2024-02-26
Add CP subsytem config for Hazelcast CR [CN-1149]

SeriyBg:
## User Impact

Add the possibility to configure CP Subsystem in the Hazelcast CR


api/v1alpha1/hazelcast_validation.go
```diff
@@ -427,3 +428,28 @@ func (v *hazelcastValidator) validateSQL(h *Hazelcast) {
...
+
+func (v *hazelcastValidator) validateCPSubsystem(h *Hazelcast) {
+	if h.Spec.CPSubsystem == nil {
+		return
+	}
+
+	cp := h.Spec.CPSubsystem
+	if h.Spec.ClusterSize == nil {
+		if cp.MemberCount > 3 {
+			v.Invalid(Path("spec", "cpSubsystem", "memberCount"), cp.MemberCount, "can not be greater the clusterSize")
```
cagric0, line 440: Shouldn't we use different error messages for different cases to distinguish the failure reason?

api/v1alpha1/hazelcast_validation.go
```diff
@@ -427,3 +428,28 @@ func (v *hazelcastValidator) validateSQL(h *Hazelcast) {
...
+	}
+
+	if cp.GroupSize != nil {
+		if (*cp.GroupSize != 3 && *cp.GroupSize != 5 && *cp.GroupSize != 7) || *cp.GroupSize > cp.MemberCount {
+			v.Invalid(Path("spec", "cpSubsystem", "memberCount"), cp.GroupSize, "can be 3, 5, or 7, but not greater that memberCount")
+		}
+	}
+
+	if cp.PVC == nil && h.Spec.Persistence.Pvc == nil {
+		v.Required(Path("spec", "cpSubsystem", "pvc"), "PVC should be configured")
```
cagric0, line 453: The point is if cp is enabled either cp.Pvc or persistence.Pvc must be configured?

api/v1alpha1/hazelcast_validation.go
```diff
@@ -427,3 +428,28 @@ func (v *hazelcastValidator) validateSQL(h *Hazelcast) {
...
+		}
+	} else if cp.MemberCount > *h.Spec.ClusterSize && *h.Spec.ClusterSize != 0 { // ClusterSize of 0 means pausing the cluster, should be ignored
+		v.Invalid(Path("spec", "cpSubsystem", "memberCount"), cp.MemberCount, "can not be greater the clusterSize")
+	}
+
+	if cp.GroupSize != nil {
+		if (*cp.GroupSize != 3 && *cp.GroupSize != 5 && *cp.GroupSize != 7) || *cp.GroupSize > cp.MemberCount {
+			v.Invalid(Path("spec", "cpSubsystem", "memberCount"), cp.GroupSize, "can be 3, 5, or 7, but not greater that memberCount")
+		}
+	}
```
cagric0, line 450: Can `GroupSize` be less than `MemberCount`?

api/v1alpha1/hazelcast_validation.go
```diff
@@ -427,3 +428,28 @@ func (v *hazelcastValidator) validateSQL(h *Hazelcast) {
 		v.Forbidden(Path("spec", "sql", "catalogPersistence"), "catalogPersistence requires Hazelcast persistence enabled")
 	}
 }
+
+func (v *hazelcastValidator) validateCPSubsystem(h *Hazelcast) {
```
cagric0, line 432: We are missing the integration tests for these validations. Could you add them?

cagric0 (changes requested):

api/v1alpha1/hazelcast_validation.go
SeriyBg: But this is the same reason. The only thing is that the `ClusterSize` is not specified, then we use the default `3` member count

api/v1alpha1/hazelcast_validation.go
SeriyBg: Yes, it can be less but not greater than

cagric0 (approved):

PR #1051, opened 2024-03-04
[CN-962] Upgrade operator-sdk version to 1.33.0

semihbkgr:


Makefile
```diff
@@ -19,14 +19,14 @@ TOOLBIN = $(shell pwd)/bin
...
 # https://github.com/operator-framework/operator-sdk/releases
-OPERATOR_SDK_VERSION ?= v1.28.1
+OPERATOR_SDK_VERSION ?= v1.33.0
 # https://github.com/kubernetes-sigs/controller-tools/releases
-CONTROLLER_GEN_VERSION ?= v0.11.3
+CONTROLLER_GEN_VERSION ?= v0.12.0
 # https://github.com/kubernetes-sigs/controller-runtime/releases
 # It is set in the go.mod file
-CONTROLLER_RUNTIME_VERSION ?= v0.13.1
+CONTROLLER_RUNTIME_VERSION ?= v0.15.0
```
kutluhanmetin, line 29: Shouldn't it be `v0.15.2`

.github/workflows/docs-release.yaml
```diff
@@ -70,7 +70,7 @@ jobs:
       - name: Set up Golang
         uses: actions/setup-go@v5
         with:
-          go-version: "1.21.0"
+          go-version: "1.20.0"
```
kutluhanmetin, line 73: I just wonder why did we have to lower go version?

Makefile
semihbkgr: I realized that this variable is not even in use.

.github/workflows/docs-release.yaml
semihbkgr: The `go` version from `operator-sdk` v1.33 is `1.20`

PR #1053, opened 2024-03-05
Compatibility test workflow

semihbkgr:
## Description

Run E2E tests on different versions of Kubernetes clusters in order to test compatibility of the operator.

PR #1054, opened 2024-03-06
[CN-632] Compatibility tests in Kind cluster

semihbkgr:
## Description

Run compatibility tests on `kind` cluster.

`GKE` cluster in #1053 is changed to `Kind`

### DIFF
--- a/.github/workflows/e2e-kind.yaml
+++ b/.github/workflows/e2e-kind.yaml
@@ -81,7 +81,7 @@ jobs:
         edition: ["os", "ee"]
     env:
       NAMESPACE: kind-test-operator-${{ matrix.edition }}
-      RELEASE_NAME: compatibility-${{ inputs.k8s-cluster-version }}-${{ matrix.edition }}
+      RELEASE_NAME: kind-${{ inputs.k8s-cluster-version }}-${{ matrix.edition }}
       CLUSTER_NAME: kind
       IMG: ${{ needs.get-image.outputs.IMG }}
     steps:
@@ -243,6 +243,6 @@ jobs:
         if: always()
         uses: actions/upload-artifact@v3
         with:
-          name: test-report-compatibility
+          name: test-report-kind
           path: allure-results/pr/
           retention-days: 5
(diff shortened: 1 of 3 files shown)


semihbkgr:
It is ready
https://github.com/hazelcast/hazelcast-platform-operator/actions/runs/8177743104

hasancelik:
@semihbkgr with this PR. you are partially reverting your previous PR:
https://github.com/hazelcast/hazelcast-platform-operator/pull/1053/files
Could you mention it in the description?

.github/workflows/e2e-kind.yaml
```diff
@@ -0,0 +1,253 @@
...
+
+on:
+  workflow_dispatch:
+  workflow_call:
+    inputs:
+      k8s-cluster-version:
+        description: "kubernetes cluster version"
+        required: false
+        type: string
+        default: "1.29"
```
cheels, line 11: What we will do after 3-4 months when 1.30 will be released? Will we update it all the time? Maybe it is worth to make it dynamically?

.github/workflows/k8s-compatibility-test.yaml
```diff
@@ -2,18 +2,99 @@ name: (E2E) K8s Compatibility Tests
...
   run-test:
-    uses: ./.github/workflows/e2e-gke.yaml
+    uses: ./.github/workflows/e2e-kind.yaml
     secrets: inherit
     name: Run E2E tests
     strategy:
       fail-fast: false
       matrix:
-        kubernetes-versions: ["1.29", "1.28", "1.27"]
+        kubernetes-versions: ["1.29", "1.28", "1.27", "1.26", "1.25"]
```
cheels, line 98: Same here, these values should be always updated manually.  What if can add the latest version here automatically?

hasancelik (approved):

.github/workflows/e2e-kind.yaml
semihbkgr: I agree, it would be good to automate this process.
I created a separate task for that: CN-1165

cheels (approved):

PR #1066, opened 2024-03-15
[CN-1163] Improve restore E2E tests

semihbkgr:

### DIFF
--- a/test/e2e/backup_restore_test.go
+++ b/test/e2e/backup_restore_test.go
@@ -22,6 +22,7 @@ import (
 
 var _ = Describe("Hazelcast CR with Persistence feature enabled", Group("backup_restore"), func() {
 	localPort := strconv.Itoa(8400 + GinkgoParallelProcess())
+
 	backupRestore := func(hazelcast *hazelcastcomv1alpha1.Hazelcast, hotBackup *hazelcastcomv1alpha1.HotBackup, useBucketConfig bool) {
 		By("creating cluster with backup enabled")
 		CreateHazelcastCR(hazelcast)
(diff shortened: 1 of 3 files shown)


test/e2e/backup_restore_test.go
```diff
@@ -505,9 +537,7 @@ var _ = Describe("Hazelcast CR with Persistence feature enabled", Group("backup_
 			Expect(k8sClient.Create(context.Background(), hotBackup)).Should(Succeed())
 			assertHotBackupSuccess(hotBackup, 1*Minute)
 
-			By("filling the cache with entries after backup")
-			fillCachePortForward(hazelcast, cache.GetDSName(), localPort, entryCount)
```
cagric0, line 509: Why did you remove filling step after backup? Shouldn't we keep it for proper testing. As far as I can understand this test is testing second put shouldn't be appear after restore. Am I missing something?

semihbkgr: It is simply a test case for checking the persistency on the caches entries: `should check cache entry persistence after HotBackup`

> As far as I can understand this test is testing second put shouldn't be appear after restore.

I thought that we don't need to include this edge case scenario in the sole test case where we check the cache persistency. We already have multiple test cases examining the behavior of restoring when hot-restart data is present.

Additionally, it's important to note that the `validateCacheEntriesPortForward` test util function is not correctly implemented due to the lack of cache API support in the Go client. Currently, it only verifies the existence of entries in the cache, ignoring the cache size and other entries assumed not to be present in the cache."

cagric0 (approved):

PR #1069, opened 2024-03-18
[CN-1167] Fix for Platform tests 

cheels:
## Description
Fixed the failing tests on Github Action with the error “Ginkgo timed out waiting for all parallel procs to report back“


test/e2e/e2e_suite_test.go
```diff
@@ -102,13 +103,17 @@ var recordedManifests = make(map[types.NamespacedName]io.Writer)
 // manifestRecorder keeps track of applied manifests
 type manifestRecorder struct {
 	client.Client
+	Mutex sync.Mutex
```
kutluhanmetin, line 106: I think we don't need to export it, and it is better give the same name with its type.

```suggestion
	mu sync.Mutex
```

test/e2e/e2e_suite_test.go
```diff
@@ -102,13 +103,17 @@ var recordedManifests = make(map[types.NamespacedName]io.Writer)
...
 }
 
 func NewManifestRecorder(client client.Client) *manifestRecorder {
 	return &manifestRecorder{Client: client}
 }
 
 func (d *manifestRecorder) Create(ctx context.Context, obj client.Object, opts ...client.CreateOption) error {
+	d.Mutex.Lock()         // Lock before accessing recordedManifests
```
kutluhanmetin, line 114: I think code itself is easy to understand, we don't need these comments.

test/e2e/helpers_test.go
```diff
@@ -185,7 +185,7 @@ func DeletePod(podName string, gracePeriod int64, lk types.NamespacedName) {
 		if err != nil {
 			log.Fatal(err)
 		}
-		time.Sleep(5 * time.Second)
+		time.Sleep(15 * time.Second)
```
kutluhanmetin, line 188: Would it be better to use a for loop instead of waiting 15 seconds and retry?

test/e2e/platform_persistence_test.go
```diff
@@ -60,15 +58,15 @@ var _ = Describe("Platform Persistence", Label("platform_persistence"), func() {
 		err := FillMapByEntryCount(ctx, hzLookupKey, true, m.MapName(), 100)
 		Expect(err).To(BeNil())
 
-		DeletePod(hazelcast.Name+"-2", 0, hzLookupKey)
-		WaitForPodReady(hazelcast.Name+"-2", hzLookupKey, 1*Minute)
+		DeletePod(hazelcast.Name+"-2", 5, hzLookupKey)
```
kutluhanmetin, line 61: Don't we want to delete pod immediately?

kutluhanmetin (commented):
Just a couple of small comments

test/e2e/platform_persistence_test.go
cheels: No it's ok for this test. It's kind of graceful restart.

test/e2e/helpers_test.go
```diff
@@ -179,19 +179,24 @@ func DeletePod(podName string, gracePeriod int64, lk types.NamespacedName) {
 		}
 		podExists := func() bool {
 			_, err := getKubernetesClientSet().CoreV1().Pods(lk.Namespace).Get(context.Background(), podName, metav1.GetOptions{})
```
kutluhanmetin, line 181: Would you assign `getKubernetesClientSet()`'s return value to a variable and reuse it?

test/e2e/helpers_test.go
```diff
@@ -179,19 +179,24 @@ func DeletePod(podName string, gracePeriod int64, lk types.NamespacedName) {
...
-			return err == nil
+			return err != nil && !errors.IsNotFound(err)
 		}
 		err := getKubernetesClientSet().CoreV1().Pods(lk.Namespace).Delete(context.Background(), podName, deleteOptions)
-		if err != nil {
+		if err != nil && !errors.IsNotFound(err) {
```
kutluhanmetin, line 185: You can use `client.IgnoreNotFound(err)` instead.

kutluhanmetin (approved):
Just minor comments, thank you for your effort!

PR #1070, opened 2024-03-18
Fix WAN replication tests

cheels:
## Description

Temporary update Hazelcast version to 5.5.0-SNAPSHOT to fix the issue in WAN tests with happening with 5.4.0-SNAPSHOT


.github/workflows/e2e-platform-wan.yaml
```diff
@@ -294,7 +294,7 @@ jobs:
 
       - name: Run Wan Test
         run: |-
-          make test-e2e NAMESPACE=src-ns RELEASE_NAME=src-ns E2E_TEST_LABELS=platform_wan REPORT_SUFFIX=ee_01 WORKFLOW_ID=wan
+          make test-e2e NAMESPACE=src-ns RELEASE_NAME=src-ns E2E_TEST_LABELS=platform_wan REPORT_SUFFIX=ee_01 WORKFLOW_ID=wan HZ_VERSION=5.5.0-SNAPSHOT
```
hasancelik, line 297: My understanding was after the thread is using 5.4-BETA-2, as I mentioned before, we should not use 5.x before the 5.11 release

hasancelik (changes requested):

.github/workflows/e2e-platform-wan.yaml
cheels: Updated to 5.4-BETA-2

hasancelik (approved):

PR #1077, opened 2024-03-25
[CN-1188] - Upgrade controller-runtime to v0.16.3

semihbkgr:
## Description

Upgrade to `operator-sdk` version `1.34.1` which uses  `controller-runtime v0.16.3` as dependency. It will fix the multi-namespace issue.


SeriyBg (approved):

Makefile
```diff
@@ -35,12 +35,17 @@ OPM_VERSION ?= v1.26.2
...
 GINKGO_VERSION ?= $(shell go list -m -f "{{.Version}}" github.com/onsi/ginkgo/v2)
 # https://github.com/kubernetes-sigs/kustomize/releases
-KUSTOMIZE_VERSION ?= v5.0.1
+KUSTOMIZE_VERSION ?= v5.2.1
 # https://github.com/helm/helm/releases
 HELM_VERSION ?= v3.10.3
 # https://github.com/mikefarah/yq/releases
 YQ_VERSION ?= v4.30.7
 
+# CONTAINER_TOOL defines the container tool to be used for building images.
```
kutluhanmetin, line 44: Why do we need this definition? Our all commands have docker prefix which makes it a bit confusing.

semihbkgr: These changes come from the `operator-sdk` migration guide. 
We don't need to conform them tough. We can revert it.

go.mod
```diff
@@ -12,8 +12,8 @@ require (
 	github.com/hashicorp/go-version v1.6.0
 	github.com/hazelcast/hazelcast-go-client v1.4.1
 	github.com/hazelcast/platform-operator-agent v0.1.20
-	github.com/onsi/ginkgo/v2 v2.9.5
-	github.com/onsi/gomega v1.27.7
+	github.com/onsi/ginkgo/v2 v2.11.0
```
kutluhanmetin, line 15: Is upgrading controller runtime force us to upgrade these as well?

Makefile
kutluhanmetin: If you agree that it makes it confusing let's revert it.

go.mod
semihbkgr: not necessarily, but operator-sdk v1.34.1 uses these versions

Makefile
```diff
@@ -19,14 +19,14 @@ TOOLBIN = $(shell pwd)/bin
...
 # https://github.com/operator-framework/operator-sdk/releases
-OPERATOR_SDK_VERSION ?= v1.33.0
+OPERATOR_SDK_VERSION ?= v1.34.1
 # https://github.com/kubernetes-sigs/controller-tools/releases
-CONTROLLER_GEN_VERSION ?= v0.12.0
+CONTROLLER_GEN_VERSION ?= v0.13.0
 # https://github.com/kubernetes-sigs/controller-runtime/releases
 # It is set in the go.mod file
-CONTROLLER_RUNTIME_VERSION ?= v0.15.0
+CONTROLLER_RUNTIME_VERSION ?= v0.16.0
```
kutluhanmetin, line 29: Should it be
```suggestion
CONTROLLER_RUNTIME_VERSION ?= v0.16.3
```

go.mod
kutluhanmetin: if we are sure that it does not break anything, it is ok.

kutluhanmetin (commented):
I left just a couple of comments.

kutluhanmetin (approved):
looks good!

PR #1030, opened 2024-02-19
[CN-1124] Expose WAN port for Member service(a.k.a service-per-pod)

cagric0:
## Description

We decided to expose default wan port(5710) for Member services like we did for discovery service. With the additional port for Member services, corresponding `hazelcastendpoints` with the type `WAN` are created directly.

### Old version

```yaml
exposeExternally:
  type: Smart
  discoveryServiceType: LoadBalancer
  memberAccess: NodePortExternalIP
```

```sh
$ k get svc
NAME             TYPE           CLUSTER-IP      EXTERNAL-IP     PORT(S)                                                       AGE
hazelcast        LoadBalancer   10.164.15.134   35.226.75.101   5701:31425/TCP,5702:30574/TCP,8081:30858/TCP,5710:31016/TCP   75s
hazelcast-0      NodePort       10.164.5.33     <none>          5701:31001/TCP                                                75s
hazelcast-1      NodePort       10.164.3.158    <none>          5701:30343/TCP                                                75s
hazelcast-2      NodePort       10.164.5.104    <none>          5701:32602/TCP                                                75s
```

```sh
$ k get hazelcastendpoints.hazelcast.com
NAME              TYPE        ADDRESS
hazelcast         Discovery   35.226.75.101:5701
hazelcast-0       Member      34.31.31.174:31001
hazelcast-1       Member      34.31.31.174:30343
hazelcast-2       Member      34.134.146.100:32602
hazelcast-wan     WAN         35.226.75.101:5710
```

### New Version

_Notice additional  `port` for `WAN` on each `Member service` and corresponding `hazelcastEndpoints`._

```yaml
  exposeExternally:
    type: Smart
    discoveryServiceType: LoadBalancer
    memberAccess: NodePortExternalIP
```

```sh
$ k get svc
NAME                             TYPE           CLUSTER-IP      EXTERNAL-IP     PORT(S)                                                       AGE
hazelcast                        LoadBalancer   10.164.15.134   35.226.75.101   5701:31425/TCP,5702:30574/TCP,8081:30858/TCP,5710:31016/TCP   75s
hazelcast-0                      NodePort       10.164.5.33     <none>          5701:31001/TCP,5710:31375/TCP                                 75s
hazelcast-1                      NodePort       10.164.3.158    <none>          5701:30343/TCP,5710:31322/TCP                                 75s
hazelcast-2                      NodePort       10.164.5.104    <none>          5701:32602/TCP,5710:32292/TCP                                 75s
```

```sh
[...]

PR #1081, opened 2024-03-28
Revert "[CN-1124] Expose WAN port for Member service(a.k.a service-per-pod)"

cagric0:
## Description

Revert #1030 

Currently k8s discovery does not accept multiple port for the member service. You can see from [here](https://github.com/hazelcast/hazelcast-mono/blob/master/hazelcast/hazelcast/src/main/java/com/hazelcast/kubernetes/KubernetesClient.java#L636)


semihbkgr (approved):
LGTM 😢

cagric0:
> LGTM, although all the tests need to be fixed

@SeriyBg  It's because of the leftover focus. I just removed it so the failures will be gone with the following run.

PR #1085, opened 2024-04-01
Added cluster_scope filtering to makefile

cheels:
## Description

Fixed cluster scope test and include only tests that will run on KIND only

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -232,6 +232,10 @@ ifeq ($(WORKFLOW_ID),pr)
 E2E_TEST_LABELS:=$(E2E_TEST_LABELS) && kind
 endif
 
+ifeq ($(WORKFLOW_ID),cluster_scope)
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -232,6 +232,10 @@ ifeq ($(WORKFLOW_ID),pr)
 E2E_TEST_LABELS:=$(E2E_TEST_LABELS) && kind
 endif
 
+ifeq ($(WORKFLOW_ID),cluster_scope)
+E2E_TEST_LABELS:=$(E2E_TEST_LABELS) && kind
```
semihbkgr, line 236: Do we really want to run `cluster_scope` tests in `kind` cluster?
or, do we move to `GKE`?

hasancelik (approved):

Makefile
semihbkgr: ok then

PR #1087, opened 2024-04-03
[CN-1118] Add Platform Tiered-Storage tests

cagric0:
## Description

Added 3 tests:
* Ensure that the cluster throws DeviceOutOfCapacityException when the data load exceeds the sum of memory-tier capacity and device capacity.
* Ensure that scaling down and up back should successfully formed the cluster and no data loss.
* Ensure that member join the cluster after ungraceful member shutdown.

### DIFF
--- a/internal/naming/constants.go
+++ b/internal/naming/constants.go
@@ -42,7 +42,7 @@ const (
 
 	OperatorName         = "hazelcast-platform-operator"
 	Hazelcast            = "hazelcast"
-	HazelcastPortName    = "hazelcast-port"
+	HazelcastPortName    = "hazelcast"
 	HazelcastStorageName = Hazelcast + "-storage"
 	HazelcastMountPath   = "/data/hazelcast"
 	TmpDirVolName        = "tmp-vol"
--- a/test/e2e/platform_tiered_storage_test.go
+++ b/test/e2e/platform_tiered_storage_test.go
@@ -2,6 +2,13 @@ package e2e
 
 import (
 	"context"
+	"fmt"
+	hzclient "github.com/hazelcast/hazelcast-platform-operator/internal/hazelcast-client"
+	"github.com/hazelcast/hazelcast-platform-operator/test"
+	"io"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
+	"k8s.io/utils/pointer"
 	"strconv"
 	. "time"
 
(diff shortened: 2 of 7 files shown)


internal/naming/constants.go
```diff
@@ -42,7 +42,7 @@ const (
 
 	OperatorName         = "hazelcast-platform-operator"
 	Hazelcast            = "hazelcast"
-	HazelcastPortName    = "hazelcast-port"
+	HazelcastPortName    = "hazelcast"
```
kutluhanmetin, line 45: Why did you make this change?

test/e2e/platform_tiered_storage_test.go
```diff
@@ -62,13 +70,191 @@ var _ = Describe("Hazelcast CR with Tiered Storage feature enabled", Group("plat
...
+			var diskSizeInMb = 1000
+			ctx := context.Background()
+
+			totalMemorySize := strconv.Itoa(totalMemorySizeInMb) + "Mi"
+			nativeMemorySize := strconv.Itoa(nativeMemorySizeInMb) + "Mi"
+			mapMemoryCapacitySize := strconv.Itoa(mapMemory) + "Mi"
+			diskSize := strconv.Itoa(diskSizeInMb) + "Mi"
+			hazelcast := hazelcastconfig.HazelcastTieredStorage(hzLookupKey, deviceName, labels)
+			hazelcast.Spec.ClusterSize = pointer.Int32(2)
+			hazelcast.Spec.ExposeExternally = &hazelcastv1alpha1.ExposeExternallyConfiguration{
```
kutluhanmetin, line 98: Do we need ExposeExternally to be enabled for Tiered Storage?

kutluhanmetin (approved):
Just small comments LGTM, thanks!

internal/naming/constants.go
cagric0: This port name is used in discovery service as well, but they were not matching. For now It doesn't affect anything but later when we change k8s discovery related configurations, it can matter.

test/e2e/platform_tiered_storage_test.go
cagric0: No we don't need. But, since maploader is used to put data, we needed to exposeExternally

cheels (approved):

cheels:
One test is failed http://reportboard.s3-website-us-east-1.amazonaws.com/aks/1842/index.html#suites/ca14ff82cab0c6d5629eb8e90a04f059/1ab685a087b0ef9/

cheels (approved):

PR #1090, opened 2024-04-05
[CN-1139] Resource limit CPU and Memory ph metrics

semihbkgr:


SeriyBg:
@semihbkgr I'm not sure we will ever need such filtering; however, since the `Quantity` struct allows easily to get the string or numeral representation, I would vote for the numeral one

SeriyBg (approved):

PR #1099, opened 2024-04-15
Update HZ and MC version to 5.4.0

cheels:
## Description

Update Hazelcast and Management Center Version to 5.4.0


semihbkgr (commented):
It is moved to #1100

PR #1108, opened 2024-04-19
Validate custom config key and yaml in WebHook

semihbkgr:
## Description

Webhook validation to check if the required `hazelcast` key exists in cm and the value is in a valid yaml structure.

### DIFF
--- a/api/v1alpha1/hazelcast_validation.go
+++ b/api/v1alpha1/hazelcast_validation.go
@@ -9,6 +9,7 @@ import (
 	"sort"
 	"strings"
 
+	"gopkg.in/yaml.v2"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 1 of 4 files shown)


SeriyBg (approved):

api/v1alpha1/hazelcast_validation.go
```diff
@@ -9,6 +9,7 @@ import (
 	"sort"
 	"strings"
 
+	"gopkg.in/yaml.v2"
```
SeriyBg, line 12: Just one nit: there is `v3` available that we are using. Pleas use it here as well

semihbkgr: besides this, should we implement similar key existence validation for the license secret?

PR #1110, opened 2024-04-22
[CN-1248]Remove reports creation for non-distribution tests

cheels:
## Description

To avoid the wrong order in report, the report creation was disabled in WF if it’s not triggered by Push event into main


cagric0:
This change will not prevent report creation for the manual runs on the `main` branch right?

cheels:
> This change will not prevent report creation for the manual runs on the `main` branch right?

I fixed it. And now report will be created only in distribution tests

cagric0 (approved):

PR #1118, opened 2024-04-26
Use bucket with restricted permission in Workload Identity test

semihbkgr:

### DIFF
--- a/test/e2e/helpers_test.go
+++ b/test/e2e/helpers_test.go
@@ -957,7 +957,6 @@ func assertHotBackupStatus(hb *hazelcastcomv1alpha1.HotBackup, s hazelcastcomv1a
 			err := k8sClient.Get(
 				context.Background(), types.NamespacedName{Name: hb.Name, Namespace: hzNamespace}, hbCheck)
 			Expect(err).ToNot(HaveOccurred())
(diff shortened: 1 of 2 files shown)


hasancelik (approved):

cagric0 (approved):

test/e2e/helpers_test.go
```diff
@@ -957,7 +957,6 @@ func assertHotBackupStatus(hb *hazelcastcomv1alpha1.HotBackup, s hazelcastcomv1a
 			err := k8sClient.Get(
 				context.Background(), types.NamespacedName{Name: hb.Name, Namespace: hzNamespace}, hbCheck)
 			Expect(err).ToNot(HaveOccurred())
-			Expect(hbCheck.Status.State).ShouldNot(Equal(hazelcastcomv1alpha1.HotBackupFailure), "Message: %v", hbCheck.Status.Message)
```
kutluhanmetin, line 960: Why do we need this change?

semihbkgr: We are checking the status of the hotbackup, out of the `eventually` block: `.Should(Equal(s))`
So we don't need it inside the `eventually` call.

kutluhanmetin (approved):

PR #1125, opened 2024-04-30
Delete EKS test clusters using 'eksctl'

semihbkgr:
## Description

Use `eksctl` to delete EKS clusters after the AWS tests.
If it fails, we can continue using the current method defined in #546

### DIFF
--- a/.github/workflows/e2e-eks.yaml
+++ b/.github/workflows/e2e-eks.yaml
@@ -321,10 +321,12 @@ jobs:
 
       - name: Delete EKS cluster
         run: |-
-          CLUSTER_NAME="operator-e2e-test-${GITHUB_SHA::8}-${{ github.run_number }}"
-          source .github/scripts/utils.sh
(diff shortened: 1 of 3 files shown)


.github/workflows/e2e-eks.yaml
```diff
@@ -321,10 +321,12 @@ jobs:
 
       - name: Delete EKS cluster
         run: |-
-          CLUSTER_NAME="operator-e2e-test-${GITHUB_SHA::8}-${{ github.run_number }}"
-          source .github/scripts/utils.sh
-          wait_for_eks_stack_deleted eksctl-$CLUSTER_NAME $DELETE_EKS_STACK_TIMEOUT_IN_MINS
```
cheels, line 326: You forgot to change it in all workflows and remove function from utils.sh.

semihbkgr: 👍 I wanted to keep them in case the change doesn't work.

cheels (approved):

SeriyBg (approved):

PR #1146, opened 2024-05-15
Use port-forward instead of ExposeExternally for CP tests [CN-1272]

SeriyBg:
## Description

CP tests are failing with the Timeout waiting for the LB to be created. To fix the failure without increasing the timeout, it is better to use the port forwarding to access the HZ member


kutluhanmetin (approved):
looks good!

hasancelik (approved):

PR #1148, opened 2024-05-16
Fix map status when MapConfig creation failed [CN-1226]

SeriyBg:
## User Impact

When MapConfig creation fails to apply, the Map status will be set to Failed

### DIFF
--- a/internal/controller/hazelcast/map_controller.go
+++ b/internal/controller/hazelcast/map_controller.go
@@ -137,8 +137,8 @@ func (r *MapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 
 	ms, err := r.ReconcileMapConfig(ctx, m, h, cl, createdBefore)
 	if err != nil {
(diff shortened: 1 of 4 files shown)


internal/controller/hazelcast/map_controller.go
```diff
@@ -137,8 +137,8 @@ func (r *MapReconciler) Reconcile(ctx context.Context, req ctrl.Request) (ctrl.R
 
 	ms, err := r.ReconcileMapConfig(ctx, m, h, cl, createdBefore)
 	if err != nil {
-		return updateMapStatus(ctx, r.Client, m, recoptions.RetryAfter(retryAfterForMap),
-			withMapState(hazelcastv1alpha1.MapPending),
```
hasancelik, line 141: @SeriyBg What about if I apply the correct configuration via the same `Map CR`? Will it be updated as `Pending` then `Success`?

hasancelik (approved):

PR #1155, opened 2024-05-21
[CN-1289] Set different cluster names for each platform in distribution tests

semihbkgr:
## Description

The distribution workflow calls `run-aks` , `run-eks` and `run-gke` jobs referencing to yaml files. All these jobs create cluster with the same name. This makes it harder to filter logs by cluster in `Grafana` board. We need to create clusters with different names to be able to filter the distribution test logs.

### DIFF
--- a/.github/workflows/e2e-platfrom-soak.yaml
+++ b/.github/workflows/e2e-platfrom-soak.yaml
@@ -164,7 +164,7 @@ jobs:
 
       - name: Update kubeconfig
         run: |-
-          CLUSTER_NAME="operator-e2e-test-${GITHUB_SHA::8}-${{ github.run_number }}"
+          CLUSTER_NAME="operator-e2e-test-soak-${GITHUB_SHA::8}-${{ github.run_number }}"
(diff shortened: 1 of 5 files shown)


.github/workflows/e2e-platfrom-soak.yaml
```diff
@@ -33,7 +33,7 @@ jobs:
       - name: Create EKS cluster
         id: create-cluster
         run: |-
-          CLUSTER_NAME="operator-e2e-test-${GITHUB_SHA::8}-${{ github.run_number }}"
+          CLUSTER_NAME="operator-e2e-test-soak-${GITHUB_SHA::8}-${{ github.run_number }}"
```
semihbkgr, line 36: @cheels Does this change affect any component other than the workflow? 
I think It shouldn't .

cheels: No, it shouldn't

cheels (approved):

SeriyBg (approved):

PR #1158, opened 2024-05-22
[CN-1196] Added tool to check breaking changes between two versions

cheels:
## Description

- Added tool that will check backward compatibility errors. It will run as part of PR builder and Test Release workflow.
- The result will be the next:

image


tools/version-diff-tool/main.go
```diff
@@ -3,295 +3,267 @@ package main
...
-	if err != nil {
-		panic(fmt.Errorf("an error occurred while downloading base bundle yaml: %s", err.Error()))
-	}
+type CRD struct {
+	APIVersion string `yaml:"apiVersion"`
+	Kind       string `yaml:"kind"`
+	Metadata   struct {
+		Name string `yaml:"name"`
+	} `yaml:"metadata"`
+	Spec struct {
```
SeriyBg, line 30: Wouldn't it be better to have a name struct for each field, like `type CrdStruct struct {` instead of anonymous ones?

SeriyBg (approved):
Looks awesome! Thanks!
Left just one small comment

.github/workflows/pull-request.yml
```diff
@@ -111,6 +111,28 @@ jobs:
...
+      - uses: actions/setup-go@v5
+        with:
+          go-version: ${{ env.GO_VERSION }}
+
+      - name: Check breaking changes between stable and latest-snapshot version
+        working-directory: tools/version-diff-tool
+        run: |
+          helm repo add hazelcast https://hazelcast-charts.s3.amazonaws.com && helm repo update
+          STABLE_VERSION=$(curl -s https://artifacthub.io/api/v1/packages/helm/hazelcast/hazelcast-platform-operator | jq -r '.available_versions | map(select(.version | contains("snapshot") | not)) | .[0].version')
+          LATEST_SNAPSHOT=$(curl -s https://artifacthub.io/api/v1/packages/helm/hazelcast/hazelcast-platform-operator | jq -r '.version')
```
kutluhanmetin, line 132: It should not be LATEST_SNAPSHOT for PR builder. In PR builder, we need to compare the changes in the PR with the STABLE_VERSION.

tools/version-diff-tool/main.go
kutluhanmetin: I thought if it could be possible to use `CustomResourceDefinition` type from `apiextensions-apiserver` dependency?

tools/version-diff-tool/main.go
```diff
@@ -3,295 +3,267 @@ package main
...
+func main() {
+	base := flag.String("base", "", "Version of the first CRD to compare")
+	revision := flag.String("revision", "", "Version of the second CRD to compare")
+	ignoreMessages := flag.String("ignore", "", "Comma-separated list of messages https://github.com/Tufin/oasdiff/blob/main/checker/localizations_src/en/messages.yaml to ignore in the output")
+	flag.Parse()
 
-	diffFile, err := formatDiff(base, rev)
-	if err != nil {
-		return "", err
+	if *base == "" || *revision == "" {
```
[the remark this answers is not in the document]
cheels: No I excluded 5.5 from the tool. It's not required anymore because we are not going to check backeward compatibility for such an old version

.github/workflows/pull-request.yml
cheels: Ok, thanks, I'll fix it

tools/version-diff-tool/main.go
```diff
@@ -3,295 +3,267 @@ package main
...
 	}
 
-	err = filepath.Walk(wd, func(path string, info os.FileInfo, err error) error {
-		if info.IsDir() {
-			return nil
-		}
-		if filepath.Ext(path) == ".yaml" {
-			os.Remove(info.Name())
+	errs := checker.CheckBackwardCompatibility(checker.GetDefaultChecks(), diffRes, operationsSources)
+	if len(errs) > 0 || len(errs) == 0 {
```
kutluhanmetin, line 255:
```suggestion
	if len(errs) >= 0 {
```

tools/version-diff-tool/main.go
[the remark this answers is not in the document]
kutluhanmetin: I know, I agree that we should exclude. But, someone using this should may not know that it is used just for helm. So adding a validation explaining "This tool support versions > 5.6". That's all.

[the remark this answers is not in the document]
kutluhanmetin: Instead of that maybe we can just add a help command explaining the reason of the tool is written for, how it works and how to use it? It may be more useful.

tools/version-diff-tool/main.go
SeriyBg: That would be nice, but the problem there is the the structure is missing `yaml` tags, so for the marshaling we would need to write some custom code, which we be even more complex that having just the custom structs

tools/version-diff-tool/main.go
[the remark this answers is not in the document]
cheels: It throws the error if user jsut run the code like "go run main.go -base 5.7 -revision 5.8 -ignore". So since it's for our internal usage it's always easy to understand the purpose of it. At least having a JIRA ticket. 

>   -base string
>         Version of the first CRD to compare
>   -ignore string
>         Comma-separated list of messages https://github.com/Tufin/oasdiff/blob/main/checker/localizations_src/en/messages.yaml to ignore in the output
>   -revision string
>         Version of the second CRD to compare

tools/version-diff-tool/main.go
cheels: I'll make struct instead of anonymous one.

SeriyBg (approved):

PR #1162, opened 2024-05-28
Fix error with MTLS client registry after operator restart [CN-1254]

SeriyBg:
## User Impact

Fix the error `failed to get MTLS client` that can occur after the operator restart

### DIFF
--- a/internal/controller/hazelcast/jetjob_controller.go
+++ b/internal/controller/hazelcast/jetjob_controller.go
@@ -3,7 +3,6 @@ package hazelcast
 import (
 	"context"
 	"encoding/json"
-	"errors"
 	"fmt"
 	"net"
 	"path"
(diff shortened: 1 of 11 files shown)


internal/mtls/client_registry.go
```diff
@@ -39,6 +40,14 @@ func (cr *httpClientRegistry) Create(ctx context.Context, kubeClient client.Clie
...
 
+func (cr *httpClientRegistry) GetOrCreate(ctx context.Context, kubeClient client.Client, ns string) (*http.Client, error) {
+	nn := types.NamespacedName{Name: n.MTLSCertSecretName, Namespace: ns}
+	if v, ok := cr.clients.Load(nn); ok {
+		return v.(*http.Client), nil
+	}
+	return cr.Create(ctx, kubeClient, ns)
+}
+
 func (cr *httpClientRegistry) Get(ns string) (*http.Client, bool) {
```
kutluhanmetin, line 51: I think we can also remove get as it should not be used anymore. Or you can make it a private method and call it from GetOrCreate.

internal/controller/hazelcast/jetjob_controller.go
```diff
@@ -228,10 +227,9 @@ func (r *JetJobReconciler) applyJetJob(ctx context.Context, job *hazelcastv1alph
 
 func (r *JetJobReconciler) downloadFile(ctx context.Context, job *hazelcastv1alpha1.JetJob, hazelcastName types.NamespacedName, jjnn types.NamespacedName, client hzclient.Client, logger logr.Logger) error {
 	g, groupCtx := errgroup.WithContext(ctx)
-	mtlsClient, ok := r.mtlsClientRegistry.Get(hazelcastName.Namespace)
-	if !ok {
-		returnErr := errors.New("failed to get MTLS client")
-		_, _ = r.updateStatus(ctx, jjnn, failedJetJobStatus(returnErr))
+	mtlsClient, err := r.mtlsClientRegistry.GetOrCreate(ctx, r.Client, hazelcastName.Namespace)
+	if err != nil {
+		_, _ = r.updateStatus(ctx, jjnn, failedJetJobStatus(err))
```
semihbkgr, line 232: We need to return error here.

SeriyBg: I did not change that since the main codebase also does not return here. It should be double-checked, and if needed, fixed in a separate PR

kutluhanmetin (approved):

semihbkgr (approved):
LGTM 👍

PR #1167, opened 2024-06-03
[CN-1271] Fix WAN replication flaky test

semihbkgr:
## Description

`Hazelcast WAN Updating WAN configuration initially fails after removal of replicated Hazelcast CR, then succeeds after removal it from the WAN spec`

The flakiness is caused by a race condition in hazelcast and wan reconciler.
- The test case fails when hazelcast is not deleted when the wan reconciler is triggered.
- The test case passes when hazelcast is deleted before the wan reconciler is triggered.

Wan reconciler should be triggered when the resource reference Hazelcast is deleted.


internal/controller/hazelcast/wanreplication_controller.go
```diff
@@ -796,3 +809,36 @@ func (r *WanReplicationReconciler) wanRequestsForTerminationCandidateMap(_ conte
 
 	return reqs
 }
+
+func (r *WanReplicationReconciler) wanRequestsForCreatedOrDeletedHazelcast(ctx context.Context, m client.Object) []reconcile.Request {
```
SeriyBg, line 813: I'm a bit unsure about the name. The `ForCreatedOrDeleted` part is not related to what the function does, but rather to how it is used.

semihbkgr: Yes 👍, you're right. I forgot to rename it.

SeriyBg: Well, the `wanRequestsForSuccessfulMap` and `wanRequestsForTerminationCandidateMap` checks the Map status for success and terminating, respectfully. There the function does not include any logic mentioned in the name

internal/controller/hazelcast/wanreplication_controller.go
```diff
@@ -729,40 +741,41 @@ func (r *WanReplicationReconciler) wanRequestsForSuccessfulMap(ctx context.Conte
...
-							Name:      wan.Name,
-							Namespace: wan.Namespace,
-						},
-					})
-					break wanResourcesLoop
+					requestedWans[wan.Name] = wan
 				}
 			}
+			if _, ok := requestedWans[wan.Name]; ok {
+				break
```
kutluhanmetin, line 763: Why we break in this line? I think we should keep iterating. I would suggest adding
`
if _, ok := requestedWans[wan.Name]; ok {
    continue
}
`
before `switch` stattement.

internal/controller/hazelcast/wanreplication_controller.go
semihbkgr: so, isn't it ok? `wanRequestsForDeletedHazelcast`

internal/controller/hazelcast/wanreplication_controller.go
semihbkgr: It iterates through the WAN items in the outer loop. Then, it iterates through the resources of each item in the inner loop. If the WAN was added as a request (`requestedWans`), we don't need to continue iterating through the following resources because it has already been added.

kutluhanmetin (approved):

SeriyBg (approved):

PR #1170, opened 2024-06-05
[CN-1283][CN-1282]: Remove OS Support

kutluhanmetin:
## Description

- Remove OS from Github workflows.
- Make `licenseKeySecretName` required.
- Add logic to hazelcast reconciler to delete if the cluster is not enterprise (using cluster authentication codec's response. In the response, if failOver is supported, it is EE otherwise OS)
- Remove `EE` and `AnyLicense` tags from tests and fix tests according to it.

## User Impact

Users won't be able to create OS Hazelcast clusters using Hazelcast Platform Operator.


internal/controller/hazelcast/hazelcast_resources.go
```diff
@@ -2954,6 +2953,30 @@ func (r *HazelcastReconciler) ensureClusterActive(ctx context.Context, client hz
 	return svc.ChangeClusterState(ctx, codecTypes.ClusterStateActive)
 }
 
+var illegalClusterType = errs.New("only enterprise clusters are supported")
+
+func (r *HazelcastReconciler) ensureClusterEnterprise(ctx context.Context, client hzclient.Client, h *hazelcastv1alpha1.Hazelcast) error {
```
SeriyBg, line 2958: I would suggest this method to return `bool, err` instead. This way, you won't have to check that the return type is `illegalClusterType`, and there will be clear separation: if an error is returned then something went wrong, otherwise check for the `bool` value

SeriyBg (commented):
Besides two comments, it would also be great to have an unit or integration test for this check

internal/controller/hazelcast/hazelcast_controller.go
```diff
@@ -264,6 +269,14 @@ func (r *HazelcastReconciler) Reconcile(ctx context.Context, req ctrl.Request) (
 	}
 	r.statusServiceRegistry.Create(req.NamespacedName, cl, r.Log, r.triggerReconcileChan)
 
+	if err = r.ensureClusterEnterprise(ctx, cl, h); errs.Is(err, illegalClusterType) {
+		err2 := r.Client.Delete(ctx, &statefulSet)
```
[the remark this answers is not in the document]
kutluhanmetin: https://github.com/hazelcast/hazelcast-platform-operator/pull/1170/commits/8bfb05b0357fb2939d614314f93e63e4d6b3e62a

internal/controller/hazelcast/hazelcast_resources.go
SeriyBg: No, as the reconciler will return the `illegalClusterType` in the reconciler itself

kutluhanmetin: https://github.com/hazelcast/hazelcast-platform-operator/pull/1170/commits/6ff3c3c7bfc47fdc7edd1520c42fbc98d9c13b95

SeriyBg (approved):

semihbkgr (approved):

PR #1172, opened 2024-06-07
[CN-1226] Fix pending state for failing map config

SeriyBg:


kutluhanmetin (commented):
As you removed pending state completely, I think we should remove other references as well. Am I wrong?

SeriyBg:
> As you removed pending state completely, I think we should remove other references as well. Am I wrong?

I haven't removed it completely, there is still one case where it's used 😄

kutluhanmetin (approved):

semihbkgr (approved):

PR #1181, opened 2024-06-24
Fix metallB setup issue

cheels:
## Description
Fix the IPAddressPool "kind-pool" is invalid: spec.addresses[0]: Invalid value: "array": spec.addresses[0] in body must be of type string: "array" issue


semihbkgr (approved):

hasancelik:
> why did you delete the soak and multi-version workflows?

@semihbkgr we created soak test to mimic cloud usage(multiple stop and restart for long run) which is not real user scenario.

for multi-version suite, we were not using it actually, it was kind of broken. we will re-evaluate and implement it again

hasancelik (approved):

PR #1184, opened 2024-06-25
Removed redundant NS creation command

cheels:
## Description
Fix the "Error from server (AlreadyExists): namespaces "test-operator-os" already exists" since "eksctl create iamserviceaccount --namespace test-operator-os" command creates namespace if it doesn't exist.

https://eksctl.io/usage/iamserviceaccounts/#:~:text=If%20the%20namespace%20doesn%27t%20exist%20already%2C%20it%20will%20be%20created.

### DIFF
--- a/.github/workflows/e2e-eks.yaml
+++ b/.github/workflows/e2e-eks.yaml
@@ -9,6 +9,7 @@ env:
   GAR_REGION: us-east1
   GAR_PROJECT: hazelcast-33
   GAR_REPO: hazelcast-platform-operator
+  NAMESPACE: test-operator-ee
 
 jobs:
   prepare-env:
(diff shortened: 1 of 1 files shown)


hasancelik (changes requested):
Could you use `${NAMESPACE}` during`iam-service-account` creation?

https://github.com/hazelcast/hazelcast-platform-operator/blob/main/.github/workflows/e2e-eks.yaml#L68-L76

cheels:
> Could you use `${NAMESPACE}` during`iam-service-account` creation?
> 
> https://github.com/hazelcast/hazelcast-platform-operator/blob/main/.github/workflows/e2e-eks.yaml#L68-L76

This is not necessary because it's not matrix in this part of workflow but then we use matrix

hasancelik (approved):

kutluhanmetin (approved):

PR #1190, opened 2024-06-30
[CN-1324]Added granular filter support from the file

cheels:
## Description

- Improved compatibility tool to read filter messages from the file. 
- Supports filter both block name (e.g. _request-property-max-set_) and specific messages like (_map.hazelcast.com removed the request property 'spec.mapStore.writeCoealescing'_)


hasancelik (approved):

tools/version-diff-tool/main.go
```diff
@@ -214,14 +214,58 @@ func formatOutput(output string) string {
...
-func ignoreChanges(output string, ignoreMessages []string) string {
+func readIgnoreMessages(filePath string) ([]string, error) {
+	file, err := os.Open(filePath)
+	if err != nil {
+		return nil, err
+	}
+	defer func(file *os.File) {
+		err = file.Close()
+		if err != nil {
+
```
kutluhanmetin, line 225: You forgot to handle the error in here.

tools/version-diff-tool/main.go
```diff
@@ -214,14 +214,58 @@ func formatOutput(output string) string {
...
+		return r == '\n' || r == '\r'
+	})
+	for i, line := range lines {
+		lines[i] = strings.TrimSpace(line)
+	}
+	return strings.Join(lines, " ")
+}
+
+func replaceANSI(str string) string {
+	return regexp.MustCompile(`\x1b\[[0-9;]*m`).ReplaceAllString(str, "")
```
kutluhanmetin, line 252: I am not a regex and ansi expert but would it be better to use `\x1b\[[0-9;]*[a-zA-Z]` ?

cheels: It's not required in this case because this regex match ANSI codes that specifically end with the letter "m" and it always ends with 'm' : https://pkg.go.dev/github.com/shiena/ansicolor#section-readme (like \x1b[33m)

kutluhanmetin: if it is sufficient I am ok with that 👍

kutluhanmetin (approved):

PR #1195, opened 2024-07-02
[CN-1329]: Add cannot be empty validation for env var name

kutluhanmetin:
## Description

This PR adds cannot be empty validation to env var name.

## User Impact

Users won't be able to create CR's which has env var name empty in it.


semihbkgr (approved):
integration tests would be great.

SeriyBg (approved):

